PR #3, opened 2015-05-13
Add RecordWriter interface

vejnar:
Similarly to RecordReader, add RecordWriter interface in order to write to SAM or BAM transparently.

The SAM writer doesn't have a Close() (BAM does), so I added one that does nothing for SAM to implement the same interface.

Please consider for merging. Thanks.
Charles

### DIFF
--- a/sam/sam.go
+++ b/sam/sam.go
@@ -129,6 +129,12 @@ type RecordReader interface {
 	Read() (*Record, error)
 }
 
+// RecordWriter wraps types that can write SAM Records.
+type RecordWriter interface {
+	Write(r *sam.Record) error
+	Close() error
+}
+
 // Iterator wraps a Reader to provide a convenient loop interface for reading SAM/BAM data.
 // Successive calls to the Next method will step through the features of the provided
 // Reader. Iteration stops unrecoverably at EOF or the first error.
@@ -205,6 +211,11 @@ func (w *Writer) Write(r *Record) error {
 	b = append(b, '\n')
 	_, err = w.w.Write(b)
 	return err
+}
+
+// Close closes the writer.
+func (w *Writer) Close() error {
+	return nil
 }
 
 const (


kortschak:
There are two issues with that that need clarifying, but I am inclined not to merge this.

1. What is the use case for this interface? `RecordReader` is included in the package because it is needed for creating an `Iterator`. If there were a type that depended on a record writing interface that would be defined here too, but it is not onerous for a client implementing such a type to include the interface definition there (this is the beauty of implicit interfaces).

2. `Close` should not be in the interface. The client provides a type that can be used for writing records. The receiving function does not now if the client wants to close at the end of the writer's use, so the call to `Close` should be left to the client.

vejnar:
1. Use case. Writing code compatible with SAM and BAM output: opening a BAM or SAM Writer used as a RecordWriter in the rest of the code.

2. I don't think it's really convenient but that could be removed. Having Close() allows to be agnostic of Writer being sam or bam.

kortschak:
I understand the abstract use cases that might exist, I'm asking for concrete use cases.

To answer my own question, I see two possible broad categories of concrete use cases.

1. A generic writer component in an application (main package).

2. A generic writer for external consumption.

In the first case what we see being written is:

```
package main

import (
	"fmt"
	"log"

	"github.com/biogo/hts/bam"
	"github.com/biogo/hts/sam"
)

type recordWriter interface {
	Write(*Record) error
}

func writeRecordsTo(w recordWriter, data interface{}) error {
	for {
		rec := process(data)
		if rec == nil {
			break
		}
		err := w.Write()
		if err != nil {
			return err
		}
	}
	return nil
}

var w = os.Stderr

func main() {
	data := getData()
	h := magicHeaderMaker()

	// write SAM
	sw := sam.NewWriter(w, h, 0)
	log.Println(writeRecordsTo(sw, data))

	// write BAM
	bw := bam.NewWriter(w, h, 0)
	log.Println(writeRecordsTo(sw, data))
	bw.Close()
}
```

No other package needs to know about the interface, but it costs only three lines. The general approach to defining interfaces is at the site of use (this is broken in a number of places, but usually with some justification somewhere - gonum/matrix/mat64 breaks this because we use the interface definitions as a design document).

If the second case, we see

```
package foo

import (
	"fmt"
	"log"

	"github.com/biogo/hts/bam"
	"github.com/biogo/hts/sam"
)

type RecordWriter interface {
	Write(*Record) error
}

func WriteRecordsTo(w RecordWriter, data interface{}) error {
	for {
		rec := process(data)
		if rec == nil {
			break
		}
		err := w.Write()
		if err != nil {
			return err
		}
	}
	return nil
}
```

In neither case was a `sam.RecordWriter` needed.

To the second issue. It might be more convenient, but it is less flexible without additional boilerplate at the call site.

In the above code, a caller could call `WriteRecordsTo` multiple times without any problem. If we add `Close() error` to the interface we are saying that the function expects something that has a `Close` method, presumably because we want to call it. Now we need to provide an additional parameter `doClose bool` to `WriteRecordsTo` to indicate whether we want to close the writer or not. It's easier to just call `Close` after return if that is what is wanted.

If you truly want a generic writer closer function, that can done (using the second case since I can't see any reason for including it in the first - though change in capitalisation gets you that):

```
func CloseWriter(w RecordWriter) error {
	if c, ok := w.(io.Closer); ok {
		return c.Close()
	}
	return nil
}
```

Alternatively, if your function want to advertise that a passed writer will be close in the signature (instead of in the documentation - which is perfectly valid), you can:

```
type RecordWriterCloser interface {
	RecordWriter
	io.Closer
}
```

and use that in the signature. Then also provide

```
type WriterCloser struct {
	RecordWriter
}

func (w WriterCloser) Close() error {
	if c, ok := w.RecordWriter.(io.Closer); ok {
		return c.Close()
	}
	return nil
}
```

The proliferation of options is an excellent argument as to why this interface should not be included in the sam package itself.

vejnar:
Thanks for the suggestions.
I implemented RecordWriter locally and called Close only for bam Writer.

PR #4, opened 2015-09-03
tabix.Index.Chunks now expects a record interface

brentp:


tabix/tabix.go
```diff
@@ -98,8 +97,13 @@ func (i *Index) Add(r Record, c bgzf.Chunk, placed, mapped bool) error {
 }
 
 // Chunks returns a []bgzf.Chunk that corresponds to the given genomic interval.
-func (i *Index) Chunks(r *sam.Reference, beg, end int) []bgzf.Chunk {
-	chunks := i.idx.Chunks(r.ID(), beg, end)
+func (i *Index) Chunks(r Record) []bgzf.Chunk {
+	refName := r.RefName()
+	id, ok := i.nameMap[refName]
+	if !ok {
+		return make([]bgzf.Chunk, 0)
```
kortschak: This should return nil.

kortschak:
While we are breaking API, how do you feel about adding an error return to this? Currently we return an empty []bgzf.Chunk, but there are two possible reasons for this - the reference is not known or the interval is outside the range of the reference. It would mean that Chunks can't be used inline, but it is probably worth doing. I'll add this if you think it's worthwhile - it touches more than just here.

brentp:
that makes sense to me.

kortschak:
OK, I've got that change ready. I just need the nit fixed and I can move on that.

brentp:
ok. done. I just sqaushed and force pushed.

kortschak:
LTGM, thanks.

PR #9, opened 2015-09-09
bgzf/index: fix repeat reading

kortschak:
@brentp Please take a look.


brentp:
LGTM assuming that test fails before the change.

kortschak:
They did.

Issue #10, opened 2015-09-11
tabix query missing bytes.


brentp:
apologies for the case, but it is reproducible. We are using a normalized version of ExAC for annotation and I ran trhough and intersected every variant with itself to make sure to get a hit. There is a single failure.

The file is here: http://s3.amazonaws.com/gemini-annotations/ExAC.r0.3.sites.vep.tidy.vcf.gz (and .tbi)

and the code is below. Run with:

```Shell
go run main.go ExAC.r0.3.sites.vep.tidy.vcf.gz
```

It is querying for the location:  `location{"X", 229379 - 2, 229382}`
And the output includes the variant of interest (X:229379), but it is appended to the end another line (X:229380)

Here is the output from the htslib tabx:

```Shell
$ tabix  /usr/local/src/gemini_install/data/gemini/data/ExAC.r0.3.sites.vep.tidy.vcf.gz X:229378-229380 | cut -f 1-5
X	229275	.	CACACGGCGACCATGGGAACCCCCTCTCCTGGGCACGTGCTCACCGCAGCTGTCGTACGGCACCACTGAGACGACAGGGACCCCCTGCCCTCCCCCGGGCGAGTCCTCACCGGTG	C
X	229379	.	CCTCACCGGTGACACGGAGACCGCGGAAGGCCCCTCCCCTGGGCGCGTG	C
X	229380	.	C	G
```

```go
package main

import (
	"bufio"
	"compress/gzip"
	"fmt"
	"io"
	"io/ioutil"
	"os"
	"strings"

	"github.com/biogo/hts/bgzf"
	"github.com/biogo/hts/bgzf/index"
	"github.com/biogo/hts/tabix"
)

func check(err error) {
	if err != nil {
		panic(err)
	}
}

type location struct {
	chrom string
	start int
	end   int
}

func (s location) RefName() string {
	return s.chrom
}
func (s location) Start() int {
	return s.start
}
func (s location) End() int {
	return s.end
}

func main() {

	path := os.Args[1]

	fh, err := os.Open(path + ".tbi")
	check(err)

	gz, err := gzip.NewReader(fh)
	check(err)
	defer gz.Close()

	idx, err := tabix.ReadTabix(gz)
	check(err)

	b, err := os.Open(path)
	check(err)
	bgz, err := bgzf.NewReader(b, 1)
	//	bgz.Blocked = false

	check(err)

	chunks, err := idx.Chunks(location{"X", 229379 - 1, 229380})
	check(err)

[...]

PR #14, opened 2015-09-15
bgzf/index: fix chunk reader iteration

kortschak:
No tests. The test needs to come, but it requires engineering a set of blocks and index chunks to mimic the output shown in #10.

@brentp Please take a look.

### DIFF
--- a/bgzf/index/index.go
+++ b/bgzf/index/index.go
@@ -50,7 +50,11 @@ func NewChunkReader(r *bgzf.Reader, chunks []bgzf.Chunk) (*ChunkReader, error) {
 
 // Read satisfies the io.Reader interface.
 func (r *ChunkReader) Read(p []byte) (int, error) {
-	if len(r.chunks) == 0 || vOffset(r.r.LastChunk().End) >= vOffset(r.chunks[0].End) {
+	if len(r.chunks) == 0 {
+		return 0, io.EOF
+	}
+	last := r.r.LastChunk()
+	if vOffset(last.End) >= vOffset(r.chunks[0].End) {
 		return 0, io.EOF
 	}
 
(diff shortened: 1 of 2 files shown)


brentp:
I don't grok why this works and why it didn't before.

kortschak:
The file offset part of a virtual offset is the unique identifier for a BGZF block. I was using the end block of the last read (which may actually be the next block - so the condition failed, and we did not truncate the []byte to the correct length - thus read potentially up to the end of the BGZF block).

The logic is still weird for me and I'm looking into it further.

kortschak:
I've improved the clarity, properly terminate in the case of non-scanner reading (it was covering for an error in my logic which is now fixed). But please run the new code on your complete set to see if you find any other cases.

PTAL

bgzf/index/index.go
```diff
@@ -58,24 +62,34 @@ func (r *ChunkReader) Read(p []byte) (int, error) {
...
 	// beyond the end of the block because the bgzf.Reader is in
 	// blocked mode and so will stop there anyway.
-	if r.r.LastChunk().End.File == r.chunks[0].End.File {
-		p = p[:min(len(p), int(r.chunks[0].End.Block-r.r.LastChunk().End.Block))]
+	// To do this we set the cursor to the end of the last read
+	// if we are still in the same block, and truncate p to the
+	// length of the remaining chunk data.
+	var cursor uint16
+	if last.Begin.File == r.chunks[0].Begin.File {
+		cursor = last.End.Block
```
brentp: If I change this to:
              cursor = r.chunks[0].Begin.Block

Then my tests (doing tons of intersections and comparing to tabix CLI tool) are getting a lot farther, but still failing eventually. I think it's in the block below for switching chunks to early??

kortschak:
PTAL

I resorted to paper and pencil after constructing tests.

brentp:
LGTM

kortschak:
Nope, not yet.

kortschak:
There is a remaining issue, though it doesn't break things, it just results in unused reads. The fix requires new API in bgzf.Reader.

PR #18, opened 2015-09-17
bgzf/index: export errors

### LINKED ISSUE #16: hts/tabix: move ErrNoReference et al out of internal
I want to check for ErroNoReference but can't do so in go 1.5 because it is in "internal". I can make a PR to move it if you have a suggestion where.

kortschak:
Fixes #16.

@brentp Please take a look.


brentp:
LGTM

PR #20, opened 2015-09-18
internal/index: fix overly aggressive optimisation

### LINKED ISSUE #19: tabix misses intervals at start of file
With this setup:

```
echo $'chr1\t1\t100' | bgzip -c > t.bed.gz; tabix t.bed.gz
```

And the script below, tabix misses all intervals. Even if I have thousands of intervals at < 10KB position, it will return empty chunks. This change seems to fix:

```Diff
diff --git a/internal/index.go b/internal/index.go
index 1287db0..baaf656 100644
--- a/internal/index.go
+++ b/internal/index.go
@@ -293,6 +293,7 @@ func OverlappingBinsFor(beg, end int) []uint32 {
        for _, r := range []struct {
                offset, shift uint32
        }{
+               {level0, level0Shift},
                {level1, level1Shift},
                {level2, level2Shift},
                {level3, level3Shift},
```

```Go
package main

import (
	"compress/gzip"
	"io/ioutil"
	"log"
	"os"

	"github.com/biogo/hts/bgzf"
	"github.com/biogo/hts/bgzf/index"
	"github.com/biogo/hts/tabix"
)

func check(err error) {
	if err != nil {
		panic(err)
	}
}

type location struct {
	chrom string
	start int
	end   int
}

func (s location) RefName() string {
	return s.chrom
}
func (s location) Start() int {
	return s.start
}
func (s location) End() int {
	return s.end
}

func main() {

	path := os.Args[1]

	fh, err := os.Open(path + ".tbi")
	check(err)

	gz, err := gzip.NewReader(fh)
	check(err)
	defer gz.Close()

	idx, err := tabix.ReadFrom(gz)
	check(err)

	b, err := os.Open(path)
	check(err)
	bgz, err := bgzf.NewReader(b, 2)

	check(err)

	chunks, err := idx.Chunks(location{"chr1", 1, 19999999})
	check(err)
	log.Println(chunks)

	cr, err := index.NewChunkReader(bgz, chunks)
	buf, _ := ioutil.ReadAll(cr)

	log.Println(len(buf))

}
```

kortschak:
Fixes #19.

@brentp Please take a look.


brentp:
LGTM. 
how do you generate the testing byte slice from the file? hexdump?

kortschak:
[utter.Dump](https://github.com/kortschak/utter) of the bytes that were read in.

PR #21, opened 2015-09-18
bgzf,bgzf/index: prevent ChunkReader overreading when an end is at block offset 0

kortschak:
@brentp Please take a look.


kortschak:
*ping*

brentp:
I didn't test this against htslib tabix as I did previous changes, but looks good.

kortschak:
I ran the tests in bix/main. It might be worth running your more extensive tests at some point. I'll merge because there are bug fixes here.

Issue #15, opened 2015-09-15
hts/tabix: add name -> id func to api


brentp:
ReferenceStats takes an int but nameMap is not exposed to the user. I guess it might be guaranteed that the index inf the Index.Names() slice is the id to send to ReferenceStats, but it would be nice to have a function like

```Go
func (i *Index) Id(name string) (id int, ok bool) {
    id, ok := i.nameMap[name]
}
```

so that the Id could then be sent to ReferenceStats

kortschak:
I would rather not. It's faster to write a helper that does this.

```
func refStatsFor(name string, idx *tabix.Index) (stats index.ReferenceStats, ok bool) {
    for i, n := range idx.Names() {
        if n == name {
            return idx.ReferenceStats(i), true
        }
    }
    return index.ReferenceStats{}, false
}
```

and not writing the method allows clients to decide whether they want the map or the iteration approach.

brentp:
that works.

PR #24, opened 2015-09-21
tabix: add method to get name -> id lookup

kortschak:
@brentp Please take a look. This is another way to address what you asked for in #15, and we already allow people to do Bad Things™, so why not add this.

### DIFF
--- a/tabix/tabix.go
+++ b/tabix/tabix.go
@@ -51,6 +51,12 @@ func (i *Index) Names() []string {
 	return i.refNames
 }
 
+// IDs returns a map of strings to integer IDs. The returned
+// map should not be altered.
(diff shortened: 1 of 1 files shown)


brentp:
you had me convinced it wasn't needed, but this seems like a sane and simple way to expose the name -> int map.

kortschak:
Yeah, I don't think it is needed, but only providing the value rather than a method makes the choice of approach much more up to the client. Giving a method takes that out of their hands (even if only psychologically).

kortschak:
Forgot to push -f. The commit here is fea1d8cc5efb92f67e667443e02cc0a893e52446.

PR #25, opened 2015-10-02
dont panic when closing a bgzf reader multiple times.

brentp:


kortschak:
What's the need for this? Note the comment here https://golang.org/pkg/io/#Closer.

(I thought similar things for another io.Reader but found I was misleading myself).

brentp:
I got around this by being more careful. But I figured panic was a harsh penalty for closing > 1x.

kortschak:
>I got around this by being more careful.

So it did its job :)

> But I figured panic was a harsh penalty for closing > 1x.

An alternative way of stating this is that a "panic is a harsh penalty for incorrect program logic."

Issue #26, opened 2016-01-06
sam: header parsing does not conform the the SAM spec


kortschak:
This is the result of incorrectly reading the spec. The reading I made was that tags were a closed set. This is not the case, and so bam and sam will not read some valid BAM/SAM files in the wild.

The fix is not immediately obvious (there are three reasonable options - none good).

1. add OtherTags map[Tag]string fields to the four types above and add values to those when the statically defined field does not exist. This is bad because there are two ways to get a tag and when a new SAM tag is promoted to predefined in the spec we cannot promote it to a static field without breaking backwards compatibility

2. add Tags map[Tag]string fields to the four types above and add all values to those even when the statically defined field does exist. This is bad because there are two places to store a record value and so there needs to be a way to reconcile conflicts.

3. move all tag record values to a map[Tag]string field in each type above and provide accessors for the required tags. This is bad because it breaks compatibility.

I think 3. is the least bad at this stage.

PR #27, opened 2016-01-19
sam: fix HD line parsing for cases where tags are user defined

kortschak:
I ended up going with the othertags approach. The publicly visible map turned out to be a lot more complicated.

@brentp Please take a look.

### DIFF
--- a/sam/header.go
+++ b/sam/header.go
@@ -94,13 +94,21 @@ type Header struct {
 	Version    string
 	SortOrder  SortOrder
 	GroupOrder GroupOrder
-	Comments   []string
+	otherTags  []tagPair
+
 	refs       []*Reference
 	rgs        []*ReadGroup
 	progs      []*Program
 	seenRefs   set
 	seenGroups set
 	seenProgs  set
+
+	Comments []string
+}
+
+type tagPair struct {
+	tag   Tag
+	value string
 }
 
 // NewHeader returns a new Header based on the given text and list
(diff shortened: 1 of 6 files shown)


kortschak:
If you think this is OK, I'll extend the logic to SQ, RG and PG lines.

brentp:
LGTM

kortschak:
Added logic and tests for other types.

Fixes #26.

PTAL

sam/header.go
```diff
@@ -126,12 +134,85 @@ func NewHeader(text []byte, r []*Reference) (*Header, error) {
...
+}
+
 // Clone returns a deep copy of the receiver.
 func (bh *Header) Clone() *Header {
 	c := &Header{
 		Version:    bh.Version,
 		SortOrder:  bh.SortOrder,
 		GroupOrder: bh.GroupOrder,
+		otherTags:  append([]tagPair(nil), bh.otherTags...),
 		Comments:   append([]string(nil), bh.Comments...),
```
brentp: I haven't seen this idiom, it just allows you to create (and append to) an empty slice in-place?

brentp:
LGTM

sam/header.go
kortschak: Use it sparingly. It's the equivalent of `func(src []T) []T { dst := make([]T, len(src)); copy(dst, src); return dst }`. rsc has expressed significant distaste for it, but here it makes sense where it's in a struct declaration and there are only two cases - with different types.

PR #29, opened 2016-01-31
sam: fix f type parsing

### LINKED ISSUE #28: sam: invalid aux tag field: sam: unknown type float64
Hi, I encountered an error message when reading sam files generated by NextGenMap:
sam: invalid aux tag field: sam: unknown type float64

I think the reason is the following code in line 176 in auxtags.go:
f, err := strconv.ParseFloat(string(tf[2]), 32)

ParseFloat will return f as float64 no matter using bit size 64 or 32, as explained in the strconv document. So the following type checking in the NewAux function will always failed.

kortschak:
Fixes #28.


kortschak:
TBR @brentp

PR #31, opened 2016-04-15
bam: add capacity to read reduced complexity records faster

### LINKED ISSUE #30: bam: consider adding Skip method
Consider adding a Skip method that reads to the next record without parsing data beyond what is necessary for this.

This allows fast record counting.

kortschak:
Fixes #30.

@brentp Please take a look. The time saving is pretty good, but I'm not sure that the API complexity addition and code growth is worth it, so I'd appreciate your view.

### DIFF
--- a/bam/bam_test.go
+++ b/bam/bam_test.go
@@ -63,32 +63,35 @@ func (s *S) TestRead(c *check.C) {
 			lines:  1000,
 		},
 	} {
-		br, err := NewReader(bytes.NewBuffer(t.in), *conc)
-		c.Assert(err, check.Equals, nil)
-
-		c.Check(br.Header().Version, check.Equals, t.header.Version)
-		c.Check(br.Header().SortOrder, check.Equals, t.header.SortOrder)
-		c.Check(br.Header().GroupOrder, check.Equals, t.header.GroupOrder)
-		c.Check(br.Header().Comments, check.DeepEquals, t.header.Comments)
-		c.Check(br.Header().Refs(), check.DeepEquals, t.header.Refs())
-		c.Check(br.Header().RGs(), check.DeepEquals, t.header.RGs())
-		c.Check(br.Header().Progs(), check.DeepEquals, t.header.Progs())
-
-		var lines int
-		for {
-			_, err := br.Read()
-			if err != nil {
-				c.Assert(err, check.Equals, io.EOF)
-				break
-			}
-			lines++
-		}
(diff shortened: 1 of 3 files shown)


brentp:
your call. the time savings is ~25% on my machine.
given the code duplication and API addition, it doesn't seem worth it. 

Spit-balling, but is there a reason to prefer this separate function from, e.g.
```
bam.SetCore(true)
core, err := bam.Read()
``

or make the sig of Read as
```
func (br *Reader) Read(core ...bool) (*sam.Record, error) {
```

then have the single Read method with a couple of extra ifs?

**Comment by brentp** [comment]
BTW, you can avoid a map lookup when `nextRefID == refID` and just set `MateRef = Ref` in those cases. That seems to give a small speedup.

**Comment by kortschak** [comment]
25% is on the order of an hour each run for the kind of analysis I'm doing. I agree the API is not nice, though I prefer a new method over API breakage or secondary knobs. I'll see if I can come up with something nicer - I imagine it will still involve the code duplication.

That optimisation is nice. Thanks for pointing it out.

**Comment by brentp** [comment]
I guess `Read()` could call `ReadCore()`

Hopefully the improvements to the gzip stuff in golang/go for v1.7 will also improve bam reading speed.

**Comment by kortschak** [comment]
They do - I run tip at the moment and I see ~30% improvement over earlier versions.

Unfortunately, there is state that is not retained in the sam.Record that is needed during the read, so doing that is not possible. A common helper might work though. I'll have a look and see what the performance impact of that is on Monday.

**Comment by kortschak** [comment]
I've been convinced of the API knob approach. Helpers are difficult and don't ease the complexity.

PTAL

**Comment by brentp** [comment]
LGTM. I had to look up why seq and auxTags had to be defined before the goto (https://github.com/golang/go/issues/8042)

In the nextRefID block, you can put the common case of refID == nextRefID before the out of range check. Probably makes no measurable difference...

Would it be over-engineering to provide constants for the omit levels?
e.g.

```golang
const (
   None = iota
   Some
   More
)
```
... with better names.

kortschak:
That issue is slightly different, but yes I tried to use a switch instead and it made the code more convoluted.

>In the nextRefID block, you can put the common case of refID == nextRefID before the out of range check.

Yes, that's cleaner.

I had actually given them names, but also a type and it seemed too heavy. Just names is good.

kortschak:
Thanks for the review.

PR #35, opened 2016-10-14
Add Cigar.RefLen()

brentp:
When parsing SA tags or MD, we may also need to know the length often
the chimeric alignment or determine the length of the Cigar. With this
change, it is possible to do so without replicating code in Record.End()

I'm not sure RefLen() is the best name...

### DIFF
--- a/sam/cigar.go
+++ b/sam/cigar.go
@@ -52,6 +52,19 @@ func (c Cigar) String() string {
 	return b.String()
 }
 
+// Lengths returns the number of reference and read bases described by the Cigar.
+func (c Cigar) Lengths() (ref, read int) {
+	var con Consume
+	for _, co := range c {
+		con = co.Type().Consumes()
+		if co.Type() != CigarBack {
+			ref += co.Len() * con.Reference
+		}
+		read += co.Len() * con.Query
+	}
+	return ref, read
+}
+
 // CigarOp is a single CIGAR operation including the operation type and the
 // length of the operation.
 type CigarOp uint32
--- a/sam/sam_test.go
+++ b/sam/sam_test.go
@@ -696,3 +696,26 @@ func (s *S) TestIssue26(c *check.C) {
 	c.Check(prog.Get(idTag), check.Equals, "program")
 	c.Check(prog.Get(fuTag), check.Equals, "bar")
 }
+
+var cigTests = []struct {
+	cig  []byte
+	ref  int
+	read int
+}{
+	{[]byte("151M"), 151, 151},
+	{[]byte("10S10M"), 10, 20},
+	{[]byte("11H11M"), 11, 11},
+	{[]byte("11H1D11M"), 12, 11},
+	{[]byte("5M21N5M"), 31, 10},
+	{[]byte("21N"), 21, 0},
+}
+
+func (s *S) TestLengths(c *check.C) {
+	for _, ct := range cigTests {
+		cig, err := ParseCigar(ct.cig)
+		c.Check(err, check.IsNil)
+		ref, read := cig.Lengths()
+		c.Check(ref, check.Equals, ct.ref)
+		c.Check(read, check.Equals, ct.read)
+	}
+}


sam/cigar.go
```diff
@@ -74,16 +88,17 @@ func (co CigarOp) String() string { return fmt.Sprintf("%d%s", co.Len(), co.Type
...
-	CigarSkipped                        // Skipped region from the reference.
-	CigarSoftClipped                    // Soft clipping (clipped sequences present in SEQ).
-	CigarHardClipped                    // Hard clipping (clipped sequences NOT present in SEQ).
-	CigarPadded                         // Padding (silent deletion from padded reference).
-	CigarEqual                          // Sequence match.
-	CigarMismatch                       // Sequence mismatch.
-	CigarBack                           // Skip backwards.
+	CigarMatch     CigarOpType = iota // Alignment match (can be a sequence match or mismatch).
+	CigarInsertion                    // Insertion to the reference.
+	CigarDeletion                     // Deletion from the reference.GATAA //  CigarDeletion       0        1
```
kortschak: Oops. Please fix this.

brentp:
thanks for the review. I updated. I don't understand this part:
```
       if con.Query != 0 {
              end = max(end, pos)
       }
```
but I used it in Lengths() as well.

sam/cigar.go
```diff
@@ -52,18 +52,19 @@ func (c Cigar) String() string {
 	return b.String()
 }
 
-// RefLen returns the total number of Reference bases consumbed by the Cigar
-func (c Cigar) RefLen() int {
-	var pos, end int
+// Lengths returns the total number of Reference, and Read bases consumbed by the Cigar
+func (c Cigar) Lengths() (int, int) {
```
kortschak: Here named returns help with documentation, so return `(ref, read int)`.

sam/cigar.go
```diff
@@ -52,18 +52,19 @@ func (c Cigar) String() string {
 	return b.String()
 }
 
-// RefLen returns the total number of Reference bases consumbed by the Cigar
-func (c Cigar) RefLen() int {
-	var pos, end int
+// Lengths returns the total number of Reference, and Read bases consumbed by the Cigar
```
kortschak: `// Lengths returns the number of reference and read bases described by the Cigar.`

sam/cigar_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import "gopkg.in/check.v1"
+
+type CigarSuite struct{}
+
+var _ = check.Suite(&CigarSuite{})
+
+var cigTests = []struct {
+	cig        []byte
+	refLength  int
```
kortschak: Just `ref` and `read` please.

sam/cigar_test.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (s *CigarSuite) TestLengths(c *check.C) {
+	for _, ct := range cigTests {
+		cig, err := ParseCigar(ct.cig)
+		c.Check(err, check.IsNil)
+		ref, read := cig.Lengths()
+		c.Check(ref, check.Equals, ct.refLength)
+		c.Check(read, check.Equals, ct.readLength)
+
```
kortschak: Delete blank line.

kortschak:
The `con.Query` test is to not do anything with the `CigarBack` operation which is poorly defined and makes no sense in many cases where lengths are wanted.

brentp:
You mean this:
```
  func (c Cigar) Lengths() (int, int) {
      var r, q int
      var con Consume
      for _, co := range c {
          con = co.Type().Consumes()
          r += co.Len() * con.Reference
          q += co.Len() * con.Query
      }
      return r, q
  }
  ```

File: sam/cigar.go [hunk 6]
```diff
@@ -76,7 +91,7 @@ type CigarOpType byte
 const (
 	CigarMatch       CigarOpType = iota // Alignment match (can be a sequence match or mismatch).
 	CigarInsertion                      // Insertion to the reference.
-	CigarDeletion                       // Deletion from the reference.
+	CigarDeletion                       // Deletion from the reference.GATAA //  CigarDeletion       0        1
```
kortschak: Still something wrong here.

kortschak:
I don't understand. You need to have the check in Lengths for query non-consumption.

kortschak:
Sorry. Yes, with a check for `co.Type() != CigarBack`, that should be OK.

sam/cigar_test.go
```diff
@@ -0,0 +1,32 @@
...
+var cigTests = []struct {
+	cig  []byte
+	ref  int
+	read int
+}{
+	{[]byte("151M"), 151, 151},
+	{[]byte("10S10M"), 10, 20},
+	{[]byte("11H11M"), 11, 11},
+	{[]byte("11H1D11M"), 12, 11},
```
kortschak: Last thing. It's probably worth having a test for a skipped (N) operation, and I think just put this test in sam_test.go

kortschak:
Please squash before you merge.

brentp:
I squashed.

PR #37, opened 2016-10-21
replace bytes.Compare with faster comparison for tag.

brentp:
Record.Tag() is showing up in my profiling. This drastically reduces the time spent there:

The time for the code below gives:
```
benchmark            old ns/op     new ns/op     delta
BenchmarkStats-4     28.0          7.36          -73.71%

benchmark            old allocs     new allocs     delta
BenchmarkStats-4     0              0              +0.00%

benchmark            old bytes     new bytes     delta
BenchmarkStats-4     0             0             +0.00%
```

```Go
package main

import (
    "log"
    "testing"

    "github.com/biogo/hts/sam"
)

func main() {
}

func mustAux(a sam.Aux, err error) sam.Aux {
    if err != nil {
        panic(err)
    }
    return a
}

func BenchmarkStats(b *testing.B) {
    r := &sam.Record{
        AuxFields: []sam.Aux{
            mustAux(sam.NewAux(sam.NewTag("NM"), "AAAAAAAAAAAAAAAA")),
            mustAux(sam.NewAux(sam.NewTag("XA"), "AAAAAAAAAAAAAAAA")),
            mustAux(sam.NewAux(sam.NewTag("MC"), "AAAAAAAAAAAAAAAA")),
            mustAux(sam.NewAux(sam.NewTag("SA"), "ref,29,-,6H5M,17,0;")),
        },
    }
    key := []byte("SA")
    b.ReportAllocs()

    for i := 0; i < b.N; i++ {
        t, ok := r.Tag(key)
        if !ok {
            log.Fatal("didn't find tag")
        }
        _ = t

    }
}
```

### DIFF
--- a/sam/record.go
+++ b/sam/record.go
@@ -105,12 +105,15 @@ func IsValidRecord(r *Record) bool {
 // Tag returns an Aux tag whose tag ID matches the first two bytes of tag and true.
 // If no tag matches, nil and false are returned.
 func (r *Record) Tag(tag []byte) (v Aux, ok bool) {
-	for i := range r.AuxFields {
-		if bytes.Compare(r.AuxFields[i][:2], tag) == 0 {
-			return r.AuxFields[i], true
+	if len(tag) < 2 {
+		panic("sam: tag too short")
+	}
+	for _, aux := range r.AuxFields {
+		if aux.matches(tag) {
+			return aux, true
 		}
 	}
-	return
+	return nil, false
 }
 
 // RefID returns the reference ID for the Record.
--- a/sam/auxtags.go
+++ b/sam/auxtags.go
@@ -461,6 +461,10 @@ func (a Aux) Value() interface{} {
 	}
 }
 
+func (a Aux) matches(tag []byte) bool {
+	return a[1] == tag[1] && a[0] == tag[0]
+}
+
 // AuxFields is a set of auxiliary fields.
 type AuxFields []Aux
 


sam/record.go
```diff
@@ -105,9 +105,12 @@ func IsValidRecord(r *Record) bool {
 // Tag returns an Aux tag whose tag ID matches the first two bytes of tag and true.
 // If no tag matches, nil and false are returned.
 func (r *Record) Tag(tag []byte) (v Aux, ok bool) {
-	for i := range r.AuxFields {
-		if bytes.Compare(r.AuxFields[i][:2], tag) == 0 {
-			return r.AuxFields[i], true
+	if len(tag) != 2 {
+		return
```
kortschak: `return nil, false` if we are going to return.

(also please fix my bad form in this below).

sam/record.go
```diff
@@ -105,9 +105,12 @@ func IsValidRecord(r *Record) bool {
 // Tag returns an Aux tag whose tag ID matches the first two bytes of tag and true.
 // If no tag matches, nil and false are returned.
 func (r *Record) Tag(tag []byte) (v Aux, ok bool) {
-	for i := range r.AuxFields {
-		if bytes.Compare(r.AuxFields[i][:2], tag) == 0 {
-			return r.AuxFields[i], true
+	if len(tag) != 2 {
```
kortschak: This should test `len(tag) < 2`. I'm not sure that we should return instead of panic here. A panic was the previous behaviour, though we could be more informative.

sam/record.go
```diff
@@ -105,9 +105,12 @@ func IsValidRecord(r *Record) bool {
...
+		return
+	}
+	for _, aux := range r.AuxFields {
+		if aux[0] == tag[0] && aux[1] == tag[1] {
```
kortschak: Can you replacing this with `if aux.matches(tag) {` adding the following method:

```
func (a Aux) matches(tag []byte) bool { return a[0] == tag[0] && a[1] == tag[1] }
```
probably in `auxtags.go` at around 463.

This will be inlined, and it improves the readability.

sam/record.go
brentp: previous behavior would return false as well.

I don't have a preference over panic or false. But, I guess in this case, panic makes sense since it should not occur for ordinary execution.

brentp: for what it's worth, doing the length check must remove some bounds checks as removing it slows the benchmark by 5-10%.

kortschak: >previous behavior would return false as well.

Only in the case for a non-match where `len(tag) >= 2`. The slice operation in `bytes.Compare` would panic for shorted tag slices.

sam/record.go
```diff
@@ -105,15 +105,15 @@ func IsValidRecord(r *Record) bool {
 // Tag returns an Aux tag whose tag ID matches the first two bytes of tag and true.
 // If no tag matches, nil and false are returned.
 func (r *Record) Tag(tag []byte) (v Aux, ok bool) {
-	if len(tag) != 2 {
-		return
+	if len(tag) < 2 {
+		return nil, false
```
kortschak: Yeah, I think a panic here: `panic("sam: tag too short")`

brentp:
added and squashed. thanks for the review.

kortschak:
Thanks.

PR #38, opened 2016-10-22
speed up bam.Read

brentp:
This was an experiment to see if it would improve bam parsing speed.
It replaces the smaller reads with a single, large read and then just works off the []byte.

With Omit(0) on an example bam, I get 15.5 seconds with this change and 18.3 on master.
With 3 cpus, I get 7.9 secs here and 11.3 master.
For Omit > 0, the difference, as expected is smaller, but still present.

The `TestConceptualBAM` test is failing. I'm not sure how to address that.

### DIFF
--- a/bam/reader.go
+++ b/bam/reader.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"unsafe"
 
 	"github.com/biogo/hts/bgzf"
@@ -50,7 +49,6 @@ func NewReader(r io.Reader, rd int) (*Reader, error) {
 		return nil, err
 	}
 	br.lastChunk.End = br.r.LastChunk().End
-
 	return br, nil
 }
 
@@ -109,90 +107,48 @@ func (br *Reader) Read() (*sam.Record, error) {
 		return nil, io.EOF
 	}
 
-	r := errReader{r: br.r}
-	bin := binaryReader{r: &r}
-
-	// Read record header data.
-	blockSize := int(bin.readInt32())
-	r.n = 0 // The blocksize field is not included in the blocksize.
-
-	// br.r.Chunk() is only valid after the call the Read(), so this
-	// must come after the first read in the record.
-	tx := br.r.Begin()
-	defer func() {
-		br.lastChunk = tx.End()
-	}()
+	b, err := newBuffer(br)
+	if err != nil {
+		return nil, err
+	}
 
 	var rec sam.Record
-
-	refID := bin.readInt32()
-	rec.Pos = int(bin.readInt32())
-	nLen := bin.readUint8()
-	rec.MapQ = bin.readUint8()
-	_ = bin.readUint16()
-	nCigar := bin.readUint16()
-	rec.Flags = sam.Flags(bin.readUint16())
-	lSeq := bin.readInt32()
-	nextRefID := bin.readInt32()
-	rec.MatePos = int(bin.readInt32())
-	rec.TempLen = int(bin.readInt32())
-	if r.err != nil {
-		return nil, r.err
-	}
-
+	refID := b.readInt32()
+	rec.Pos = int(b.readUint32())
+	nLen := b.readUint8()
+	rec.MapQ = b.readUint8()
+	b.discard(2)
+	nCigar := b.readUint16()
+	rec.Flags = sam.Flags(b.readUint16())
+	lSeq := int32(b.readUint32())
+	nextRefID := int32(b.readUint32())
+	rec.MatePos = int(b.readInt32())
+	rec.TempLen = int(b.readInt32())
 	// Read variable length data.
-	name := make([]byte, nLen)
-	if nf, _ := r.Read(name); nf != int(nLen) {
-		return nil, errors.New("bam: truncated record name")
-	}
-	rec.Name = string(name[:len(name)-1]) // The BAM spec indicates name is null terminated.
+	rec.Name = string(b.bytes(int(nLen) - 1))
+	b.discard(1)
 
-	rec.Cigar = readCigarOps(&bin, nCigar)
-	if r.err != nil {
-		return nil, r.err
-	}
+	rec.Cigar = readCigarOps(b.bytes(int(nCigar) * 4))
 
 	var seq doublets
 	var auxTags []byte
 	if br.omit >= 2 {
 		goto done
 	}
-	seq = make(doublets, (lSeq+1)>>1)
-	if nf, _ := r.Read(seq.Bytes()); nf != int((lSeq+1)>>1) {
-		return nil, errors.New("bam: truncated sequence")
-	}
-	rec.Seq = sam.Seq{Length: int(lSeq), Seq: seq}
 
-	rec.Qual = make([]byte, lSeq)
-	if nf, _ := r.Read(rec.Qual); nf != int(lSeq) {
-		return nil, errors.New("bam: truncated quality")
-	}
+	seq = make(doublets, (lSeq+1)>>1)
+	*(*[]byte)(unsafe.Pointer(&seq)) = b.bytes(int(lSeq+1) >> 1)
+
+	rec.Seq = sam.Seq{Length: int(lSeq), Seq: seq}
+	rec.Qual = b.bytes(int(lSeq))
 
 	if br.omit >= 1 {
 		goto done
 	}
-	auxTags = make([]byte, blockSize-r.n)
-	r.Read(auxTags)
-	if r.n != blockSize {
-		return nil, errors.New("bam: truncated auxiliary data")
-	}
+	auxTags = b.bytes(b.len())
 	rec.AuxFields = parseAux(auxTags)
 
-	if r.err != nil && r.err != io.EOF {
-		return nil, r.err
-	}
-
 done:
-	if br.omit > 0 {
-		// Discard unused record data.
-		_, err := io.CopyN(ioutil.Discard, &r, int64(blockSize-r.n))
-		if err != nil {
-			return nil, errors.New("bam: truncated record")
-		}
-		if r.err != nil {
-			return nil, r.err
-		}
-	}
 	refs := int32(len(br.h.Refs()))
 	if refID != -1 {
 		if refID < -1 || refID >= refs {
@@ -323,13 +279,11 @@ func (i *Iterator) Close() error {
 	return i.Error()
 }
 
-func readCigarOps(br *binaryReader, n uint16) []sam.CigarOp {
-	co := make([]sam.CigarOp, n)
+// len(cb) must be a multiple of 4.
+func readCigarOps(cb []byte) []sam.CigarOp {
+	co := make([]sam.CigarOp, len(cb)/4)
 	for i := range co {
-		co[i] = sam.CigarOp(br.readUint32())
-		if br.r.err != nil {
-			return nil
-		}
+		co[i] = sam.CigarOp(binary.LittleEndian.Uint32(cb[i*4 : (i+1)*4]))
 	}
 	return co
 }
@@ -386,6 +340,71 @@ func parseAux(aux []byte) (aa []sam.Aux) {
 	return
 }
 
+// buffer is light-weight read buffer.
+type buffer struct {
+	off  int
+	data []byte
+}
+
+func (b *buffer) bytes(n int) []byte {
+	s := b.off
+	b.off += n
+	return b.data[s:b.off]
+}
+
+func (b *buffer) len() int {
+	return len(b.data) - b.off
+}
+
+func (b *buffer) discard(n int) {
+	b.off += n
+}
+
+func (b *buffer) readUint8() uint8 {
+	b.off++
+	return b.data[b.off-1]
+}
+
+func (b *buffer) readUint16() uint16 {
+	return binary.LittleEndian.Uint16(b.bytes(2))
+}
+
+func (b *buffer) readInt32() int32 {
+	return int32(binary.LittleEndian.Uint32(b.bytes(4)))
+}
+
+func (b *buffer) readUint32() uint32 {
+	return binary.LittleEndian.Uint32(b.bytes(4))
+}
+
+// newBuffer returns a new buffer reading from the Reader's underlying bgzf.Reader and
+// updates the Reader's lastChunk field.
+func newBuffer(br *Reader) (*buffer, error) {
+	var buf [4]byte
+	n, err := io.ReadFull(br.r, buf[:4])
+	// br.r.Chunk() is only valid after the call the Read(), so this
+	// must come after the first read in the record.
+	tx := br.r.Begin()
+	if err != nil {
+		return nil, err
+	}
+	if n != 4 {
+		return nil, errors.New("bam: invalid record: short block size")
+	}
+	b := &buffer{data: buf[:4]}
+	size := int(b.readInt32())
+	b.off, b.data = 0, make([]byte, size)
+	n, err = io.ReadFull(br.r, b.data)
+	br.lastChunk = tx.End()
+	if err != nil {
+		return nil, err
+	}
+	if n != size {
+		return nil, errors.New("bam: truncated record")
+	}
+	return b, nil
+}
+
 // buildAux constructs a single byte slice that represents a slice of sam.Aux.
 func buildAux(aa []sam.Aux) (aux []byte) {
 	for _, a := range aa {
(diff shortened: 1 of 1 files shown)


kortschak:
I can't accept this until the conceptual bam test passes. That test behaviour that is necessary for BAM indexing.

kortschak:
I don't think this is the right way to do it.

What you are doing is intended to do a buffered read of the record. This work all ends up in the `Read` method, but it is not necessary for this to work. All you need to do is add a buffer to `errReader` that you can preload with `block_size` bytes from the `bgzf.Reader` after you have read the first four bytes. This will leave the body of `Read` essentially unchanged (the nature of errors will be difference - unfortunately losing the context information content of the errors).

There are also other potential target for optimisation: `binReader` should may be improved by not having methods on the pointer and there may be a benefit from using a `sync.Pool` to keep the readers (thus saving the buffer allocations).

brentp:
I did like this approach because it is transparent and even removes lots of code. Would you consider it if the tests pass?

Otherwise, you mean

```
type errReader struct {
    r   *bgzf.Reader
    n   int
    err error
    buf [bgzf.MaxBlockSize]
}
```
?
and to Read from that and return it to a sync.Pool?

kortschak:
The change moves code from a relatively regular and simple section into the spec-dependent section. This makes `Read` more difficult to vet without significantly improving the readability overall.

Yes, essentially like that, though the buffer does not need to be BGZF block size large. Probably a buffer of 4096 would be fine - a `bytes.Buffer` would work except that it reads ahead and that would break indexing. So essentially you need to mimic a `bytes.Buffer` with a read-ahead constraint.

brentp:
I gave this a try quickly here: https://github.com/brentp/hts/commit/b46df5abaecff103ff59c5b953d2cbab682e27e8
using a limited reader, a [1024]byte buffer in errReader, and a sync.Pool to reuse. 

On a bam where no blocks exceed 256. I get:

```
limited-reader
real	0m9.610s
user	0m25.300s
sys	0m0.340s

master
real	0m11.319s
user	0m27.040s
sys	0m0.344s

one-read
real	0m7.849s
user	0m24.388s
sys	0m0.392s
```

Also, the limited-reader branch still fails the ConceptualBam test.

kortschak:
That looks like it's heading in the right direction, though pools should be shared across all values in the package, not just be a single reader (there should be a helper for get and put to make sure the state is correct - see how it's done in gonum/matrix/mat64 for the types there).

There is unnecessary copying in the code as it stands (this is because of what I have asked you to do), but this can go away later if the errReader is made to be more specialised and be able to return binary decodings directly. I suspect that after that we will see similar performance as with one-read.

However, the indexing behaviour needs to be fixed.

brentp:
do you have any insight on why the TestConceptualBam is failing or how to fix? I see that on master, it never reaches the final block, instead, it sends an EOF *with* the penultimate block.

In master, it reads data and fills the sam record for the block, but then checks r.err right after parseAux() [here](https://github.com/biogo/hts/blob/50da7d4131a3b5c9d063932461cab4d1fafb20b0/bam/reader.go#L181). And it returns nil, r.err there, even though there is (what appears to be) valid data in the sam.Record.


IIUC, the reason it gets an EOF there is because `len(auxTags) == 0` and it tries to read 0 bytes. Which causes it to skip to the next bgzf block. but that still discards a valid record. In short, I think the test is flawed.

kortschak:
I think you are right.

I'll send a PR to fix that.

Did you notice that your PR panics on 1.3 and 1.4? I'm happy to drop support for those Go versions, but do you know why?

brentp:
hmm, I d/l'ed 1.4.3 to try it out. If I replace:
```
c.raw[r.off:r.plus(4)]
```
with 
```
v := c.raw[r.off:r.off + 4]
r.off += 4
```
thenblob/master/doc/go1.5.html it works in 1.4. so must be an order of operations change. I can change the read* methods to be:
```
s, e := r.off, r.plus(n)
c.raw[s:e]
```
what do you think?

kortschak:
TBH I'd like the plus API to go away. I'd prefer if you had a something that returned the bytes and updated the position it would be nicer. Then there is no need to worry about this.

brentp:
something like this: 
```
func (c *cursor) get(n int) []byte {
    s, e := c.off, c.off + n
    c.off = e
    return c.raw[s:e]
 }
```

kortschak:
Sure, though s/get/bytes/.

brentp:
Done. Any other concerns?

now the tests are failing with `gofmt` not found.

kortschak:
That's not gofmt not found - that warning always happens. However, your code is not gofmt'd.

I'd like to get #39 merged first and then rebase this on top of that as a single commit if that's OK. Then I'll have a more careful look here.

brentp:
rebased and ready for review.

bam/reader.go
```diff
@@ -101,6 +99,52 @@ const (
...
+
+func (c *cursor) readInt32() int32 {
+	return int32(binary.LittleEndian.Uint32(c.bytes(4)))
+}
+
+func (c *cursor) readUint32() uint32 {
+	return binary.LittleEndian.Uint32(c.bytes(4))
+}
+
+func newcursor(r *bgzf.Reader, buf [4]byte) (*cursor, error) {
```
kortschak: If you make this return a bgzf.Tx, then you don't need to take a [4]byte.

Rename to newBuffer when the type name changes.

bam/reader.go
```diff
@@ -101,6 +99,52 @@ const (
...
+	c := &cursor{raw: buf[:4]}
+	size := int(c.readInt32())
+	c.off, c.raw = 0, make([]byte, size)
+	n, err := r.Read(c.raw)
+	if err != nil {
+		return nil, err
+	}
+	if n != size {
```
kortschak: Move this up above the error check and make the conditional block just be an assignment to err.

bam/reader.go
```diff
@@ -109,90 +153,64 @@ func (br *Reader) Read() (*sam.Record, error) {
...
 	defer func() {
 		br.lastChunk = tx.End()
 	}()
 
-	var rec sam.Record
+	if err != nil {
+		return nil, err
+	}
+	if n != 4 {
+		return nil, errors.New("bam: truncated record")
```
kortschak: I go so far as to say invalid record at this point: "bam: invalid record: short block size".

bam/reader.go
```diff
@@ -109,90 +153,64 @@ func (br *Reader) Read() (*sam.Record, error) {
...
 	tx := br.r.Begin()
 	defer func() {
 		br.lastChunk = tx.End()
 	}()
 
-	var rec sam.Record
+	if err != nil {
+		return nil, err
+	}
+	if n != 4 {
```
kortschak: Move this above the error check and make it an assignment to err. Then move them both above the transaction handling.

bam/reader.go
```diff
@@ -109,90 +153,64 @@ func (br *Reader) Read() (*sam.Record, error) {
...
+	if err != nil {
+		return nil, err
 	}
 
+	var rec sam.Record
+	refID := c.readInt32()
+	rec.Pos = int(c.readUint32())
+	nLen := c.readUint8()
+	rec.MapQ = c.readUint8()
+	c.off += 2 // discard 2 bytes
```
kortschak: Add a discard(n int) method that does this instead of a comment.

bam/reader.go
```diff
@@ -109,90 +153,64 @@ func (br *Reader) Read() (*sam.Record, error) {
...
+	rec.MatePos = int(c.readInt32())
+	rec.TempLen = int(c.readInt32())
 	// Read variable length data.
-	name := make([]byte, nLen)
-	if nf, _ := r.Read(name); nf != int(nLen) {
-		return nil, errors.New("bam: truncated record name")
-	}
-	rec.Name = string(name[:len(name)-1]) // The BAM spec indicates name is null terminated.
+	rec.Name = string(c.bytes(int(nLen) - 1))
+	c.off++ // discard null byte
```
kortschak: Use discard(1).

bam/reader.go
```diff
@@ -109,90 +153,64 @@ func (br *Reader) Read() (*sam.Record, error) {
...
 
 	if br.omit >= 1 {
 		goto done
 	}
-	auxTags = make([]byte, blockSize-r.n)
-	r.Read(auxTags)
-	if r.n != blockSize {
-		return nil, errors.New("bam: truncated auxiliary data")
-	}
+	auxTags = c.bytes(-1)
```
kortschak: Please don't use negative counts as reverse indexing.

bam/reader.go
```diff
@@ -323,13 +341,10 @@ func (i *Iterator) Close() error {
 	return i.Error()
 }
 
-func readCigarOps(br *binaryReader, n uint16) []sam.CigarOp {
-	co := make([]sam.CigarOp, n)
+func readCigarOps(cb []byte) []sam.CigarOp {
```
kortschak: Add a comment that len(cb) must be an integral multiple of 4.

bam/reader.go
brentp: what do you mean here? something like:

```
      if _, err := r.Read(b.data); err != nil {
          return nil, err
      }
```
but then where to check the n == size ?

kortschak: Fleshing it out, it makes it worse. Ignore both of these (this one and [here](https://github.com/biogo/hts/pull/38/files/6fd6756ce6ffa0c469b7e0df2fa00c3cadcebaea#diff-79ef3a5e756c991412daca300726aa6a)). Sorry.

brentp:
I pushed a change that makes all other changes except for using bgzf.Tx which I'm not familiar with.
I will squash after you take a look.

bam/reader.go
```diff
@@ -101,6 +99,59 @@ const (
...
+	data []byte
+}
+
+func (b *buffer) bytes(n int) []byte {
+	s := b.off
+	b.off += n
+	return b.data[s:b.off]
+}
+
+func (b *buffer) rest() []byte {
```
kortschak: I'd prefer it if we did something like

```
func (b *buffer) len() int {
    return len(b.data)-b.off
}
```

and then `rec.AuxFields = parseAux(b.bytes(b.len()))` where you are reading the tags.

bam/reader.go
```diff
@@ -109,90 +160,63 @@ func (br *Reader) Read() (*sam.Record, error) {
...
 
-	r := errReader{r: br.r}
-	bin := binaryReader{r: &r}
-
-	// Read record header data.
-	blockSize := int(bin.readInt32())
-	r.n = 0 // The blocksize field is not included in the blocksize.
-
+	var buf [4]byte
+	n, err := br.r.Read(buf[:4])
```
kortschak: We need to check that this always works (the io.Reader interface doesn't guarantee it, but I'm pretty sure bgzf.Reader behaves correctly). This can happen latter, and can just be a replacement of this with a call to io.ReadFull.

bam/reader.go
```diff
@@ -109,90 +160,63 @@ func (br *Reader) Read() (*sam.Record, error) {
...
-	_ = bin.readUint16()
-	nCigar := bin.readUint16()
-	rec.Flags = sam.Flags(bin.readUint16())
-	lSeq := bin.readInt32()
-	nextRefID := bin.readInt32()
-	rec.MatePos = int(bin.readInt32())
-	rec.TempLen = int(bin.readInt32())
-	if r.err != nil {
-		return nil, r.err
+	b, err := newBuffer(br.r, buf)
```
kortschak: I was wrong before about returning the bgzf.Tx. It is even better. Since the entire record has been read, the lastChunk field can be populated in newBuffer, eliding the need for the defer.

brentp:
current state is right around 8 seconds for the same test I posted before with master at 11.3

bam/reader.go
brentp: I'll make this change. Same for the subsequent read as well?

kortschak:
Yes, this looks OK.

For the Tx stuff, you end up with a newBuffer that looks like this (untested):

```
func newBuffer(r *bgzf.Reader) (*buffer, bgzf.Chunk, error) {
	var buf [4]byte
	b := &buffer{data: buf[:4]}
	size := int(b.readInt32())
	tx := r.Begin()
	b.off, b.data = 0, make([]byte, size)
	n, err := r.Read(b.data)
	chunk := tx.End()
	if err != nil {
		return nil, chunk, err
	}
	if n != size {
		return nil, chunk, errors.New("bam: truncated record")
	}
	return b, chunk, nil
}
```

It may be better to have the [4]byte as part of the buffer, but we can try that later.

brentp:
I have this:
```
  func newBuffer(br *Reader, buf [4]byte) (*buffer, error) {
  
      b := &buffer{data: buf[:4]}
      size := int(b.readInt32())
      tx := br.r.Begin()
      b.off, b.data = 0, make([]byte, size)
      n, err := br.r.Read(b.data)
      br.lastChunk = tx.End()
      if err != nil {
          return nil, err
      }
      if n != size {
          return nil, errors.New("bam: truncated record")
      }
      return b, nil
  }
```

bam/reader.go
```diff
@@ -138,11 +136,14 @@ func (b *buffer) readUint32() uint32 {
 	return binary.LittleEndian.Uint32(b.bytes(4))
 }
 
-func newBuffer(r *bgzf.Reader, buf [4]byte) (*buffer, error) {
+func newBuffer(br *Reader, buf [4]byte) (*buffer, error) {
+
```
kortschak: Delete this line.

bam/reader.go
```diff
@@ -138,11 +136,14 @@ func (b *buffer) readUint32() uint32 {
 	return binary.LittleEndian.Uint32(b.bytes(4))
 }
 
-func newBuffer(r *bgzf.Reader, buf [4]byte) (*buffer, error) {
+func newBuffer(br *Reader, buf [4]byte) (*buffer, error) {
```
kortschak: There's no need to pass the buf in.

Can you add a doc comment:
```
// newBuffer returns a new buffer reading from the Reader's underlying bgzf.Reader and
// updates the Reader's lastChunk field.
```

brentp: The `buf` is needed as holds that bytes that indicate the block size. I could move that part out of Read() and into newBuffer. Actually, I think that would be nicer.

kortschak: *A* buf is needed, but you don't need to pass it in. Just use one off the stack.

kortschak: Oh yes. Sorry.

bam/reader.go
```diff
@@ -109,90 +161,58 @@ func (br *Reader) Read() (*sam.Record, error) {
...
-
+	var buf [4]byte
+	n, err := io.ReadFull(br.r, buf[:4])
+	if err != nil {
+		return nil, err
+	}
+	if n != 4 {
+		return nil, errors.New("bam: invalid record: short block size")
+	}
 	// br.r.Chunk() is only valid after the call the Read(), so this
```
kortschak: This comment needs to be moved with the `br.r.Begin()` call. You can fix it when you move the size read into `newBuffer`.

brentp: done.

bam/reader.go
kortschak: Sorry, above the func decl and add a period.

bam/reader.go
```diff
@@ -101,98 +99,121 @@ const (
...
 )
 
-// Read returns the next sam.Record in the BAM stream. The sam.Record
-// returned will not contain the sequence, quality or auxiliary tag
-// data if CoreOnly(true) has been called.
-func (br *Reader) Read() (*sam.Record, error) {
-	if br.c != nil && vOffset(br.r.LastChunk().End) >= vOffset(br.c.End) {
-		return nil, io.EOF
-	}
+// buffer is light-weight read buffer.
```
kortschak: Can all the buffer implementation go below Reader (where the errReader etc were). The convention is that you lead with major moving parts and follow with helpers.

bam/reader.go
```diff
@@ -101,98 +99,121 @@ const (
...
+	}
+	if n != 4 {
+		return nil, errors.New("bam: invalid record: short block size")
+	}
+	b := &buffer{data: buf[:4]}
+	size := int(b.readInt32())
+	b.off, b.data = 0, make([]byte, size)
 	// br.r.Chunk() is only valid after the call the Read(), so this
 	// must come after the first read in the record.
 	tx := br.r.Begin()
```
kortschak: Move this and the comment to just after the io.ReadFull call.

brentp: I made the changes except this one. Not sure what you mean. The tx has to span the Read() of the record and we want to check for errors before starting the tx, right?

kortschak: Just after the previous io.ReadFull. That way, the chunk at least shows where the read started if there is an error.

brentp: done

bam/reader.go
```diff
@@ -386,63 +340,82 @@ func parseAux(aux []byte) (aa []sam.Aux) {
...
+		return nil, err
+	}
+	if n != 4 {
+		return nil, errors.New("bam: invalid record: short block size")
+	}
+	b := &buffer{data: buf[:4]}
+	size := int(b.readInt32())
+	b.off, b.data = 0, make([]byte, size)
+	n, err = io.ReadFull(br.r, b.data)
+	br.lastChunk = tx.End()
```
kortschak: For this to work as I was suggesting, it does need to go into a defer, but I've put you through enough pain. So I'll fix that.

brentp: ah. that's why I was hesitant to move it. 
thanks for the review.

kortschak:
Thank you.

brentp:
In actual tests, I see 30-35% improvement from this. [A benchmark](https://github.com/brentp/biogo-bench) shows even better improvement:

```
brentp@x1:~/go/src/github.com/brentp/biogo-bench$ benchcmp before.txt after.txt 
benchmark           old ns/op     new ns/op     delta
BenchmarkRead-4     346           135           -60.98%

BenchmarkRead-4     3              1              -66.67%

benchmark           old bytes     new bytes     delta
BenchmarkRead-4     224           4             -98.21%
```

where before is d3530d4

brentp:
[here's](https://github.com/brentp/biogo-bench) a comparison to d3530d4

```
benchmark           old ns/op     new ns/op     delta
BenchmarkRead-4     8711209       6306822       -27.60%

BenchmarkRead-4     40688          34018          -16.39%

benchmark           old bytes     new bytes     delta
BenchmarkRead-4     4076490       4388706       +7.66%
```

PR #39, opened 2016-11-02
bam: fix EOF handling in Read

kortschak:
@brentp Please take a look.

The last EOF-returned bgzf.Chunk troubles me - I don't really understand why it steps back by 4 bytes (and I don't have time to really resolve this). I have skirted the issue by saying that it's not valid. If you want to find out why this happens that would be great.


brentp:
I don't really grok this. But you might as well check the last block if err != nil.

The reason it steps back for bytes is because of https://github.com/biogo/hts/blob/master/bgzf/reader.go#L699

kortschak:
If I check the last block (obviously it passes since its altered to do this), it constrains the return. Since it's not valid, there is no reason to check it.

That doesn't explain why it steps back as far as I can see. The query I have is why the bgzf.Reader has stepped back 4 bytes. The End.Block == 157 makes sense, the Begin.Block == 153 does not. Ahhh, I have it now. There was not a successful read, so the bgzf.Reader's lastChunk is what is used. I guess that is OK, but it would be nice to clean that up so that it returns a zero-length chunk if the read was zero-length with an error.

kortschak:
Nope, just thought more and I'm still confused.

kortschak:
Wrong button.

kortschak:
I can fix the issue by adding a chunk update [here](https://github.com/biogo/hts/blob/master/bgzf/reader.go#L509), or changing the documentation [here](https://github.com/biogo/hts/blob/master/bgzf/reader.go#L486) to say "successful read operation". Either is fine.

PR #42, opened 2016-11-04
re-use the [4]byte for getting blocksize.

brentp:
this gives about a 2.5% speed bump.

```
benchmark           old ns/op     new ns/op     delta
BenchmarkRead-4     6454267       6272807       -2.81%

BenchmarkRead-4     34020          30684          -9.81%
```

### DIFF
--- a/bam/reader.go
+++ b/bam/reader.go
@@ -28,6 +28,9 @@ type Reader struct {
 	omit int
 
 	lastChunk bgzf.Chunk
+
+	// buf is used to read the block size of each record.
+	buf [4]byte
 }
 
 // NewReader returns a new Reader using the given io.Reader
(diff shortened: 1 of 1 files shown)


bam/reader.go
```diff
@@ -28,6 +28,9 @@ type Reader struct {
 	omit int
 
 	lastChunk bgzf.Chunk
+
+	// used to read the block size
```
kortschak: Please use complete sentences.

brentp:
done.

bam/reader.go
```diff
@@ -28,6 +28,9 @@ type Reader struct {
 	omit int
 
 	lastChunk bgzf.Chunk
+
+	// buf is used to read the block size of each record
```
kortschak: :) With period.

PR #45, opened 2016-11-05
bam: protect aux tags from append overwrites by user

kortschak:
@brentp Please take a look.

Looking at parseAux, I think we should make that return an error rather than panicking, and be more forgiving with indexing errors. At the moment it's brittle for badly formed bam files.

### DIFF
--- a/bam/reader.go
+++ b/bam/reader.go
@@ -317,7 +317,7 @@ func parseAux(aux []byte) []sam.Aux {
 		switch j := jumps[t]; {
 		case j > 0:
 			j += 3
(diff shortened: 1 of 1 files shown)


brentp:
this is to ensure that if the user appends they get a new slice?

that seems extremely defensive, but otherwise LGTM.

kortschak:
>this is to ensure that if the user appends they get a new slice?

Yes.

>that seems extremely defensive, but otherwise LGTM.

You would be surprised what some people will do. It's better to prevent a failure than try to explain why something was a bad idea when they raise an issue.

PR #48, opened 2016-11-21
sam: fix integer range and binary Aux encodings

kortschak:
I don't know how this went unnoticed for so long.

@brentp Please take a look.


brentp:
LGTM

PR #51, opened 2016-12-03
bgzf: fix magic block detection

### LINKED ISSUE #50: bgzf tests failing on go version go1.8beta1 linux/amd64
I can also get an error to occur in 1.7.3 on the same test if I add a Flush()
That may be a different error.

```Patch
diff --git a/bgzf/bgzf_test.go b/bgzf/bgzf_test.go
index 683d69e..7543460 100644
--- a/bgzf/bgzf_test.go
+++ b/bgzf/bgzf_test.go
@@ -155,6 +155,7 @@ func TestRoundTrip(t *testing.T) {
        if _, err := w.Write([]byte("payload")); err != nil {
                t.Fatalf("Write: %v", err)
        }
+       fmt.Println(w.Flush())
        if err := w.Close(); err != nil {
                t.Fatalf("Writer.Close: %v", err)
        }
```

Here is the error for 1.8beta:

$ go test
--- FAIL: TestRoundTrip (0.00s)
	bgzf_test.go:186: comment is "", want "comment"
	bgzf_test.go:195: mtime is -62135596800, want 100000000
	bgzf_test.go:198: name is "", want "name"
FAIL
FAIL	github.com/biogo/hts/bgzf	0.137s
```

kortschak:
@brentp Please take a look.

This handles both 1.8 compress/gzip behaviour and pre-1.8 behaviour.

Also fix incorrect test error report value.

Fixes #50.

### DIFF
--- a/bgzf/bgzf_test.go
+++ b/bgzf/bgzf_test.go
@@ -186,7 +186,7 @@ func TestRoundTrip(t *testing.T) {
 		t.Errorf("comment is %q, want %q", r.Comment, "comment")
 	}
 	if bl := ExpectedMemberSize(r.Header); bl != len(MagicBlock) {
-		t.Errorf("expectedMemberSize is %d, want %d", bl, wbl)
+		t.Errorf("expectedMemberSize is %d, want %d", bl, len(MagicBlock))
 	}
 	if string(r.Extra) != "BC\x02\x00\x1b\x00" {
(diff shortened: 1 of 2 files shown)


brentp:
LGTM. should we add `tip` to .travis.yml?

kortschak:
I'll add go1.8beta1 to `.travis.yml`.

PR #55, opened 2016-12-24
internal/index: batch calls to binary.Read to limit reflection overhead

brentp:
gives about a 1.5-4X speed improvement in bam.ReadIndex, depending
on the index.

### DIFF
--- a/bam/bam_test.go
+++ b/bam/bam_test.go
@@ -47,8 +47,9 @@ var _ = check.Suite(&S{})
 const maxInt = int(^uint(0) >> 1)
 
 var (
-	file = flag.String("bench.file", "", "file to read for benchmarking")
-	conc = flag.Int("conc", 1, "sets the level of concurrency for compression")
+	file   = flag.String("bench.file", "", "file to read for benchmarking")
+	findex = flag.String("bench.index", "", "index file to read for benchmarking")
+	conc   = flag.Int("conc", 1, "sets the level of concurrency for compression")
 )
 
 func (s *S) TestRead(c *check.C) {
@@ -191,6 +192,26 @@ func BenchmarkRead(b *testing.B) {
 		}
 	}
 	f.Close()
+}
+
+func BenchmarkReadIndex(b *testing.B) {
+	if *findex == "" {
+		b.Skip("no index file specified")
+	}
+	buf, err := ioutil.ReadFile(*findex)
+	if err != nil {
+		b.Fatalf("Index read failed: %v", err)
+	}
+	r := bytes.NewReader(buf)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		r.Seek(0, os.SEEK_SET)
+		_, err = ReadIndex(r)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
 }
 
 func BenchmarkReadCoreAndSeq(b *testing.B) {
--- a/internal/index_read.go
+++ b/internal/index_read.go
@@ -106,22 +106,16 @@ func readChunks(r io.Reader, n int32, typ string) ([]bgzf.Chunk, error) {
 	if n == 0 {
 		return nil, nil
 	}
-	var (
-		vOff uint64
-		err  error
-	)
 	chunks := make([]bgzf.Chunk, n)
+	var buf [16]byte
 	for i := range chunks {
-		err = binary.Read(r, binary.LittleEndian, &vOff)
+		// Get the begin and end offset in a single read.
+		_, err := io.ReadFull(r, buf[:])
 		if err != nil {
-			return nil, fmt.Errorf("%s: failed to read chunk begin virtual offset: %v", typ, err)
+			return nil, fmt.Errorf("%s: failed to read chunk virtual offset: %v", typ, err)
 		}
-		chunks[i].Begin = makeOffset(vOff)
-		err = binary.Read(r, binary.LittleEndian, &vOff)
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to read chunk end virtual offset: %v", typ, err)
-		}
-		chunks[i].End = makeOffset(vOff)
+		chunks[i].Begin = makeOffset(binary.LittleEndian.Uint64(buf[:8]))
+		chunks[i].End = makeOffset(binary.LittleEndian.Uint64(buf[8:]))
 	}
 	if !sort.IsSorted(byBeginOffset(chunks)) {
 		sort.Sort(byBeginOffset(chunks))
@@ -165,17 +159,30 @@ func readIntervals(r io.Reader, typ string) ([]bgzf.Offset, error) {
 	if n == 0 {
 		return nil, nil
 	}
-	var vOff uint64
 	offsets := make([]bgzf.Offset, n)
-	for i := range offsets {
-		err := binary.Read(r, binary.LittleEndian, &vOff)
+	// chunkSize determines the number of offsets consumed by each binary.Read.
+	const chunkSize = 512
+	var vOffs [chunkSize]uint64
+	for i := 0; i < int(n); i += chunkSize {
+		l := min(int(n)-i, len(vOffs))
+		err = binary.Read(r, binary.LittleEndian, vOffs[:l])
 		if err != nil {
 			return nil, fmt.Errorf("%s: failed to read tile interval virtual offset: %v", typ, err)
 		}
-		offsets[i] = makeOffset(vOff)
+		for k := 0; k < l; k++ {
+			offsets[i+k] = makeOffset(vOffs[k])
+		}
 	}
+
 	if !sort.IsSorted(byVirtOffset(offsets)) {
 		sort.Sort(byVirtOffset(offsets))
 	}
 	return offsets, nil
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
 }


brentp:
BTW, it's showing that I have permission to merge this PR before travis has even finished.

bam/bam_test.go
```diff
@@ -193,6 +194,27 @@ func BenchmarkRead(b *testing.B) {
...
+	}
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		f, err := os.Open(*findex)
+		if err != nil {
+			b.Fatalf("Open failed: %v", err)
+		}
+		b.StartTimer()
+
+		_, err = ReadIndex(f)
```
kortschak: Read the file once into the bytes.Buffer and then pass the bytes to ReadIndex through a bytes.Reader. This way, you don't benchmark file reading.

internal/index_read.go
```diff
@@ -102,26 +103,19 @@ func readBins(r io.Reader, typ string) ([]Bin, *ReferenceStats, error) {
 	return bins, stats, nil
 }
 
-func readChunks(r io.Reader, n int32, typ string) ([]bgzf.Chunk, error) {
+func readChunks(r io.Reader, n int32, typ string, buf [16]byte) ([]bgzf.Chunk, error) {
```
kortschak: I don't understand this. You are stack allocating a `[16]byte` for each call by this, why not just have `var buf [16]byte` prior to the loop?

brentp:
here's the benchcmp, these are pretty stable:
```
benchmark                old ns/op     new ns/op     delta
BenchmarkReadIndex-4     7282756       3993347       -45.17%
```

and for another: 
```
benchmark                old ns/op     new ns/op     delta
BenchmarkReadIndex-4     1864271       332785        -82.15%
```

brentp:
After the change to bench:
```
benchmark                old ns/op     new ns/op     delta
BenchmarkReadIndex-4     1504545       860009        -42.84%
```

```
benchmark                old ns/op     new ns/op     delta
BenchmarkReadIndex-4     343927        177406        -48.42%
```

internal/index_read.go
brentp: I did that first but the loop is usually only run between 1 and 3 times per function call so the benefit was minor. This gave a little bit better improvement, though most of the change improvement comes from the linear index.

brentp: I changed it to a slice and get the same performance. I can push that if you prefer.

bam/bam_test.go
```diff
@@ -193,6 +194,34 @@ func BenchmarkRead(b *testing.B) {
 	f.Close()
 }
 
+func BenchmarkReadIndex(b *testing.B) {
+	if *findex == "" {
+		b.Skip("no index file specified")
+	}
+	b.StopTimer()
```
kortschak: You don't need a start/stop, you can just use a `b.ResetTimer()`.

bam/bam_test.go
```diff
@@ -193,6 +194,34 @@ func BenchmarkRead(b *testing.B) {
...
+	f, err := os.Open(*findex)
```
kortschak: Instead of all this, just use `b, err := ioutil.ReadFile(*findex)`.

bam/bam_test.go
```diff
@@ -193,6 +194,34 @@ func BenchmarkRead(b *testing.B) {
...
+	buf, err := ioutil.ReadAll(f)
+	if err != nil {
+		b.Fatalf("Index read failed: %v", err)
+	}
+	f.Close()
+
+	b.StartTimer()
+	for i := 0; i < b.N; i++ {
+
+		rdr := bytes.NewBuffer(buf)
```
kortschak: Instead of a bytes.Buffer in the loop, have a bytes.Reader that is seek(0, io.SeekStart)ed on each loop.

Please remove the blanks around the seek line.

internal/index_read.go
kortschak: If this code stays the way it is, it needs a comment explaining why; semantically it makes no sense since the buf is passed in zeroed and is copied each call (when the holiday is over, I might ask why the performance difference is seen).

What is the performance difference between `var buf [16]byte` and as it is currently?

Have you profiled why the cost is different?

kortschak:
Please repeat the benchmarks and use benchstat instead of benchcmp. Also, I'm wondering how much this actually impacts on a normal use. The index is read once and is generally fairly short. Do we actually need to do this?

brentp:
Here is the benchmark with benchstats:

```
name         old time/op  new time/op  delta
ReadIndex-4   345µs ± 1%   178µs ± 3%  -48.53%  (p=0.008 n=5+5)
```

I have changed to to allocate the buffer inside of readChunks instead of passing it in as there is no speed difference between those 2 modes.

I am building an SV viewer for [bigly](https://github.com/brentp/bigly) and this helps it to be more responsive. It halves the time to read the index.

kortschak:
>I am building an SV viewer for bigly and this helps it to be more responsive. It halves the time to read the index.

Yeah, I can understand that, I'm just wondering how much that is of the entire cost.

brentp:
> Yeah, I can understand that, I'm just wondering how much that is of the entire cost.

Up to as must as one-half

kortschak:
Really? So you're only reading index files then?

brentp:
I think I have this all cleaned up. 

> Really? So you're only reading index files then?

I'm reading the index, and then parsing the bam file and running bigly pileup on a region. If that region is small enough then reading the index can be half of the total request time. This is true for a WGS bam index where the index itself can be quite large.
And, to clarify, I mean that the time to read the index was up to half of the total request time.

internal/index_read.go
```diff
@@ -103,14 +102,15 @@ func readBins(r io.Reader, typ string) ([]Bin, *ReferenceStats, error) {
 	return bins, stats, nil
 }
 
-func readChunks(r io.Reader, n int32, typ string, buf [16]byte) ([]bgzf.Chunk, error) {
+func readChunks(r io.Reader, n int32, typ string) ([]bgzf.Chunk, error) {
 	if n == 0 {
 		return nil, nil
 	}
 	chunks := make([]bgzf.Chunk, n)
+	buf := make([]byte, 16)
```
kortschak: If this is really hitting you, you should benchmark this approach against `var buf [16]byte` and passing `buf[:]` to the `io.ReadFull` call below. Post the results from benchstat.

bam/bam_test.go
```diff
@@ -198,23 +198,15 @@ func BenchmarkReadIndex(b *testing.B) {
...
-		b.Fatalf("Open failed: %v", err)
-	}
-
-	buf, err := ioutil.ReadAll(f)
+	buf, err := ioutil.ReadFile(*findex)
 	if err != nil {
 		b.Fatalf("Index read failed: %v", err)
 	}
-	f.Close()
+	rdr := bytes.NewReader(buf)
```
kortschak: Sorry, but can this be `r`. We only live for 3 lines, so we know what it is by then.

bam/bam_test.go
```diff
@@ -198,23 +198,15 @@ func BenchmarkReadIndex(b *testing.B) {
...
 
-	b.StartTimer()
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-
-		rdr := bytes.NewBuffer(buf)
-
+		rdr.Seek(0, os.SEEK_SET)
```
kortschak: Do we support Go versions that do not have `io.SeekStart`? It's a nicer constant.

internal/index_read.go
brentp: It's not. I was mistaken about passing it in.

bam/bam_test.go
brentp: I guess that's only defined since 1.7: https://golang.org/doc/go1.7#io

brentp:
>> And, to clarify, I mean that the time to read the index was up to half of the total request time.

> OK, I'm asking about the total time spent to get from the input data to the desired result, which should include the BAM reading time and your analysis.

I haven't benchmarked that rigorously, just know that, to read the index of an example was on the order of a second and the entire request was on the order of 2 seconds

internal/index_read.go
kortschak: No, I mean on this line, use `var buf [16]byte` and then pass `buf[:]` to `io.ReadFull` below. Does that make a difference - it shouldn't, but that is easier to read.

brentp: OK, I changed to `[16]byte`. There's no difference in performance.

internal/index_read.go
```diff
@@ -106,22 +106,16 @@ func readChunks(r io.Reader, n int32, typ string) ([]bgzf.Chunk, error) {
...
 	chunks := make([]bgzf.Chunk, n)
+	var buf [16]byte
 	for i := range chunks {
-		err = binary.Read(r, binary.LittleEndian, &vOff)
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to read chunk begin virtual offset: %v", typ, err)
-		}
-		chunks[i].Begin = makeOffset(vOff)
-		err = binary.Read(r, binary.LittleEndian, &vOff)
+		// get the begin and end offset in a single read
```
kortschak: Initial capital and terminal full stop.

internal/index_read.go
```diff
@@ -156,6 +150,9 @@ func readStats(r io.Reader, typ string) (*ReferenceStats, error) {
 	return &stats, nil
 }
 
+// chunkSize determines the number of offsets  consumed by each binary.Read.
```
kortschak: `s/  / /` and this all can be put in `readIntervals`.

brentp: so you mean to define chunkSize inside of `readIntervals` as well?

kortschak: As well? It's only used there AFAICS, so it doesn't need to be here at all.

Also, `vOffs` can be declared as `var vOffs [chunkSize]uint64` below.

brentp: I moved it inside and squashed. vOffs must be a slice so it can be truncated on the final loop.

internal/index_read.go
```diff
@@ -165,15 +159,24 @@ func readIntervals(r io.Reader, typ string) ([]bgzf.Offset, error) {
...
-		err := binary.Read(r, binary.LittleEndian, &vOff)
+	// chunkSize determines the number of offsets consumed by each binary.Read.
+	chunkSize := 512
+	vOffs := make([]uint64, chunkSize)
+	for i := 0; i < int(n); i += chunkSize {
+		// Limit the size of vOffs on final iteration.
+		if i+chunkSize >= int(n) {
+			vOffs = vOffs[:int(n)-i]
+		}
+		err := binary.Read(r, binary.LittleEndian, &vOffs)
```
kortschak: `err = binary.Read(r, binary.LittleEndian, vOffs[:min(int(n)-1, chunkSize)])` and define `min(int, int) int`\*. Then the condition above goes away, you can use the `[chunkSize]uint64` value and readability improves.

Note that because `vOffs` is a slice in the existing code and the proposed code here, the address operator is not required.

\*  `min(int, int) int`
```
func min(a, b int) int {
    if a < b {
        return a
    }
    return b
}
```

brentp: I made these changes

internal/index_read.go
```diff
@@ -156,6 +150,13 @@ func readStats(r io.Reader, typ string) (*ReferenceStats, error) {
 	return &stats, nil
 }
 
+func min(a, b int) int {
```
kortschak: The convention is that helper appear after their use, or if shared in a collection of helpers. Please place this below readIntervals.

internal/index_read.go
```diff
@@ -165,15 +166,22 @@ func readIntervals(r io.Reader, typ string) ([]bgzf.Offset, error) {
...
 	offsets := make([]bgzf.Offset, n)
-	for i := range offsets {
-		err := binary.Read(r, binary.LittleEndian, &vOff)
+	// chunkSize determines the number of offsets consumed by each binary.Read.
+	const chunkSize = 512
+	var vOffs [chunkSize]uint64
+	for i := 0; i < int(n); i += chunkSize {
+		// Limit the size of vOffs on final iteration.
+		l := min(int(n)-i, chunkSize)
+		err := binary.Read(r, binary.LittleEndian, vOffs[:l])
```
kortschak: `l` doesn't add anything to readability here and is not reused, so do the calculation inline.

Also `err` is already declared, so we don't need `:=`.

The comment does not add anything since the `min(n, len(arr))` which is functionally what the expression is idiomatic for this behaviour (maybe the expression could be changed to `min(int(n)-1, len(vOffs))` to reflect this - note that the assembly that is generated should be identical since `len([nt]T)` is a compile time constant).

brentp: I made those changes except for removing `l` because that's used in the following loop.

kortschak: Sorry. Missed that use.

kortschak:
Thank you.

PR #56, opened 2016-12-31
Read full

brentp:
This gives
```
name    old time/op    new time/op    delta
Read-4    28.4ms ± 1%    26.6ms ± 1%   -6.04%  (p=0.000 n=15+12)

name    old alloc/op   new alloc/op   delta
Read-4    6.90MB ± 0%    0.21MB ± 0%  -96.99%  (p=0.000 n=15+13)

name    old allocs/op  new allocs/op  delta
Read-4       414 ± 0%       316 ± 0%  -23.67%  (p=0.000 n=15+15)
```

and with -conc 3:

```
name    old time/op    new time/op    delta
Read-4    15.5ms ± 1%    14.6ms ± 2%   -5.79%  (p=0.000 n=14+14)

name    old alloc/op   new alloc/op   delta
Read-4    10.8MB ± 0%     4.1MB ± 0%  -61.87%  (p=0.000 n=15+15)

name    old allocs/op  new allocs/op  delta
Read-4       481 ± 0%       383 ± 0%  -20.43%  (p=0.000 n=15+14)
```

in [vcfanno](https://github.com/brentp/vcfanno) that does streaming intersection of many files, this gives ~6-8% improvement in run-time.

### DIFF
--- a/bgzf/bgzf_test.go
+++ b/bgzf/bgzf_test.go
@@ -27,7 +27,10 @@ import (
 	"github.com/biogo/hts/bgzf/cache"
 )
 
-var conc = flag.Int("conc", 1, "sets the level of concurrency for compression")
+var (
+	conc = flag.Int("conc", 1, "sets the level of concurrency for compression")
+	file = flag.String("bench.file", "", "bgzf file to read for benchmarking decompression")
+)
 
 type countWriter struct {
 	bytes int64
@@ -1032,5 +1035,37 @@ func BenchmarkWrite(b *testing.B) {
 			bg.Write(block)
 		}
 		bg.Wait()
+	}
+}
+
+func BenchmarkRead(b *testing.B) {
+	if *file == "" {
+		b.Skip("no bgzf file specified")
+	}
+	f, err := os.Open(*file)
+	if err != nil {
+		b.Fatalf("file open failed: %v", err)
+	}
+	defer f.Close()
+
+	buf := make([]byte, 16384)
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		f.Seek(0, os.SEEK_SET)
+		bg, err := NewReader(f, *conc)
+		if err != nil {
+			b.Fatalf("bgzf open failed: %v", err)
+		}
+		for {
+			_, err = bg.Read(buf)
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				b.Fatalf("bgzf read failed: %v", err)
+			}
+		}
+		bg.Close()
 	}
 }
--- a/bgzf/cache.go
+++ b/bgzf/cache.go
@@ -125,15 +125,41 @@ func (b *block) Read(p []byte) (int, error) {
 	return n, err
 }
 
+// readToEOF will exhaust r or fill buf.
+// If r does not EOF upon reading up to len(buf) bytes then readToEOF will return
+// io.ErrShortBuffer and an additional byte will be discarded from the reader.
+func readToEOF(r io.Reader, buf []byte) (n int, err error) {
+	for err == nil && n < len(buf) {
+		var nn int
+		nn, err = r.Read(buf[n:])
+		n += nn
+	}
+	switch {
+	case err == io.EOF:
+		return n, nil
+	case n == MaxBlockSize && err == nil:
+		// This is paranoic, but some readers will return
+		// quickly when passed a zero-length byte slice.
+		var dummy [1]byte
+		_, err = r.Read(dummy[:])
+		if err == nil {
+			return n, io.ErrShortBuffer
+		}
+		if err == io.EOF {
+			err = nil
+		}
+	}
+	return n, err
+}
+
 func (b *block) readFrom(r io.ReadCloser) error {
 	o := b.owner
 	b.owner = nil
-	buf := bytes.NewBuffer(b.data[:0])
-	_, err := io.Copy(buf, r)
+	n, err := readToEOF(r, b.data[:])
 	if err != nil {
 		return err
 	}
-	b.buf = bytes.NewReader(buf.Bytes())
+	b.buf = bytes.NewReader(b.data[:n])
 	b.owner = o
 	b.magic = b.magic && b.len() == 0
 	return r.Close()


kortschak:
I don't like the use of `io.ReadFull` here. It is intended to signify that the full slice should be filled. The dropping of `io.UnexpectedEOF` is a code smell that follows from its use. If you want to do it this way, write a simplified copy of `io.ReadAtLeast` that just fills the slice and returns n. However, I'd like to know what is causing the iterated read operation to be faster that the `io.Copy` call. This would require profiling.

brentp:
I can change io.ReadFull to loop over `Read()` until EOF if you prefer, though I don't see why that's preferable. Would a comment above that line indicating the intent of `ReadFull` be sufficient?

Re profiling. Here is readFrom before:
```
(pprof) list readFrom
Total: 331.92MB
ROUTINE ======================== github.com/biogo/hts/bgzf.(*block).readFrom in /home/brentp/go/src/github.com/biogo/hts/bgzf/cache.go
         0   323.96MB (flat, cum) 97.60% of Total
         .          .    127:
         .          .    128:func (b *block) readFrom(r io.ReadCloser) error {
         .          .    129:	o := b.owner
         .          .    130:	b.owner = nil
         .          .    131:	buf := bytes.NewBuffer(b.data[:0])
         .   323.96MB    132:	_, err := io.Copy(buf, r)
         .          .    133:	if err != nil {
         .          .    134:		return err
         .          .    135:	}
         .          .    136:	b.buf = bytes.NewReader(buf.Bytes())
         .          .    137:	b.owner = o
```

and after, that change:
```
(pprof) list readFrom
Total: 2.68MB
```
I'm not sure why io.Copy is allocating so much, I guess it could be re-sizing the slice (not the underlying array) along with the buffer used by io.Copy. Though using io.CopyBuffer with a buffer from a sync.Pool did not improve things as much as this.

brentp:
On a 3GB VCF, here is memory profile before:
```
(pprof) top10
52.19GB of 52.40GB total (99.59%)
Dropped 14 nodes (cum <= 0.26GB)
      flat  flat%   sum%        cum   cum%
   52.14GB 99.49% 99.49%    52.14GB 99.49%  bytes.makeSlice
    0.05GB 0.098% 99.59%    52.40GB   100%  github.com/biogo/hts/bgzf.(*block).readFrom
         0     0% 99.59%    52.35GB 99.90%  bytes.(*Buffer).ReadFrom
         0     0% 99.59%    52.40GB   100%  github.com/biogo/hts/bgzf.(*decompressor).nextBlockAt.func1
         0     0% 99.59%    52.35GB 99.90%  io.Copy
         0     0% 99.59%    52.35GB 99.90%  io.copyBuffer
         0     0% 99.59%    52.40GB   100%  runtime.goexit
```
and after:
```
(pprof) top10
210.52MB of 210.52MB total (  100%)
Showing top 10 nodes out of 19 (cum >= 3MB)
      flat  flat%   sum%        cum   cum%
  193.02MB 91.69% 91.69%   193.02MB 91.69%  compress/flate.(*huffmanDecoder).init
   14.50MB  6.89% 98.57%   207.52MB 98.57%  github.com/biogo/hts/bgzf.(*block).readFrom
       3MB  1.43%   100%        3MB  1.43%  compress/gzip.(*Reader).readHeader
         0     0%   100%   193.02MB 91.69%  compress/flate.(*decompressor).Read
         0     0%   100%   193.02MB 91.69%  compress/flate.(*decompressor).nextBlock
         0     0%   100%   193.02MB 91.69%  compress/flate.(*decompressor).readHuffman
         0     0%   100%   193.02MB 91.69%  compress/gzip.(*Reader).Read
         0     0%   100%        3MB  1.43%  compress/gzip.(*Reader).Reset
         0     0%   100%        3MB  1.43%  github.com/biogo/hts/bgzf.(*Reader).Read
         0     0%   100%        3MB  1.43%  github.com/biogo/hts/bgzf.(*Reader).nextBlock
```

kortschak:
Can you look at the profiles in 'io.Copy`? Mainly who is calling makeslice so much.

brentp:
`bytes.Buffer` implements `ReadFrom` so that is used and that causes the extra allocations. See the graphs below. Also, those and the profiles above are for `-alloc_space`, when using `-alloc_objects` I see a 7% decrease in allocations which matches the performance improvement.
Here is the profile of `ReadFrom`
```
(pprof) list ReadFrom
Total: 52.27GB
ROUTINE ======================== bytes.(*Buffer).ReadFrom in /home/brentp/go/go1.8beta1/go/src/bytes/buffer.go
         0    52.21GB (flat, cum) 99.88% of Total
         .          .    168:			// not enough space at end
         .          .    169:			newBuf := b.buf
         .          .    170:			if b.off+free < MinRead {
         .          .    171:				// not enough space using beginning of buffer;
         .          .    172:				// double buffer capacity
         .    52.01GB    173:				newBuf = makeSlice(2*cap(b.buf) + MinRead)
         .          .    174:			}
         .          .    175:			copy(newBuf, b.buf[b.off:])
         .          .    176:			b.buf = newBuf[:len(b.buf)-b.off]
         .          .    177:			b.off = 0
         .          .    178:		}
         .   208.02MB    179:		m, e := r.Read(b.buf[len(b.buf):cap(b.buf)])
         .          .    180:		b.buf = b.buf[0 : len(b.buf)+m]
         .          .    181:		n += int64(m)
         .          .    182:		if e == io.EOF {
         .          .    183:			break
         .          .    184:		}
```

[before.pdf](https://github.com/biogo/hts/files/680324/before.pdf)
[after.pdf](https://github.com/biogo/hts/files/680323/after.pdf)

brentp:
[This](https://github.com/golang/go/issues/16474) suggest something to try with using anonymous structs, but that seems even uglier than ReadFull.

kortschak:
I would suggest either a read loop or use bytes.Buffer's ReadFrom method. In either case there needs to be a check for expected size. In the case of the read loop there needs to be a non progression check when the buffer slice is full (the reader still has bytes to read) and with ReadFrom there needs to be a check that the bytes.Buffer's slice has not been grown.

brentp:
I added a helper function that does the loop.

bgzf/cache.go
```diff
@@ -125,13 +126,26 @@ func (b *block) Read(p []byte) (int, error) {
...
+// If one of those is not true, the err will be non-nil.
+func readToEOF(r io.Reader, buf []byte) (n int, err error) {
+	for err == nil && n < len(buf) {
+		var nn int
+		nn, err = r.Read(buf[n:])
+		n += nn
+	}
+	if err == io.EOF {
+		err = nil
+	} else if err == nil && n != MaxBlockSize {
```
kortschak: In don't understand this. n can only be `0 <= n && n <= MaxBlockSize`. So this branch can never have a nil error according to the loop condition.

What you want to do is

```
switch {
case err == io.EOF:
    return n, nil
case n == MaxBlockSize && err == nil:
    // This is paranoic, but some readers will return
    // quickly when passed a zero-length byte slice.
    var dummy [1]byte
    _, err = r.Read(dummy[:])
    if err == nil {
        return n, io.ErrShortBuffer
    }
    if err = io.EOF {
        err = nil
    }
}
return n, err
```

Add to the comment, readToEOF may consume more bytes from the reader than returned in buf if the length of the available data is greater than MaxBlockSize, or words to that effect.

brentp:
Done.

This is somehow faster. I guess because it can't inline calls with io.Reader args and there were a lot of those now replaced by the single function. On my 3GB test file with -conc 1:
```
benchmark           old ns/op        new ns/op        delta
BenchmarkRead-4     124945818850     107597436845     -13.88%

BenchmarkRead-4     5728424        5324950        -7.04%

benchmark           old bytes       new bytes     delta
BenchmarkRead-4     56388514360     232078760     -99.59%
```
and -conc 3

```
benchmark           old ns/op       new ns/op       delta
BenchmarkRead-4     71698951874     62600825624     -12.69%

BenchmarkRead-4     6135310        5728673        -6.63%

benchmark           old bytes       new bytes       delta
BenchmarkRead-4     86095419064     29939133992     -65.23%
```

bgzf/cache.go
```diff
@@ -126,20 +125,30 @@ func (b *block) Read(p []byte) (int, error) {
...
 	}
-	if err == io.EOF {
-		err = nil
-	} else if err == nil && n != MaxBlockSize {
-		return n, fmt.Errorf("bgzf: expected EOF at end of block after reading %d bytes", n)
+	switch {
+	case err == io.EOF:
+		return n, nil
+	case n == MaxBlockSize && err == nil:
+		// some readers may return quickly when passed a zero-length byte slice.
```
kortschak: The complete text/formating here would be preferable - and even if not, with a leading capital and short over two lines. It is paranoic - the io.Reader docs say that the following call to Read should return a `0, nil`, but I know some that don't.

bgzf/bgzf_test.go
```diff
@@ -1034,3 +1037,37 @@ func BenchmarkWrite(b *testing.B) {
 		bg.Wait()
 	}
 }
+
+func BenchmarkRead(b *testing.B) {
+	if *file == "" {
+		b.Skip("no bgzf file specified")
+	}
+	fh, err := os.Open(*file)
```
kortschak: `f` is good enough. I don't think we use the term file handle anywhere.

bgzf/bgzf_test.go
```diff
@@ -1034,3 +1037,37 @@ func BenchmarkWrite(b *testing.B) {
...
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				b.Fatalf("bgzf read failed: %v", err)
+			}
+		}
+		bgz.Close()
+	}
+	b.ReportAllocs()
```
kortschak: Please remove this; you can get allocation reporting during a benchmark by `-test.benchmem`.

bgzf/bgzf_test.go
```diff
@@ -1034,3 +1037,37 @@ func BenchmarkWrite(b *testing.B) {
...
+	}
+	defer fh.Close()
+
+	buf := make([]byte, 16384)
+	b.ResetTimer()
+	//b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		fh.Seek(0, os.SEEK_SET)
+		bgz, err := NewReader(fh, *conc)
```
kortschak: s/bgz/bg/g please.

brentp:
addressed and squashed

kortschak:
Thank you.

Issue #57, opened 2017-01-05
bgzf output not compatible with htslib


brentp:
if I bgzip a bed file with biogo, it's not usable from tabix. Here's a test that demonstrates the problem:

```Go
func TestHeader(t *testing.T) {
    buf := new(bytes.Buffer)
    gz := NewWriter(buf, 1)
    gz.Comment = "comment"
    gz.Extra = []byte("extra")

    _, err := gz.Write([]byte("payload"))
    if err != nil {
        t.Fatal("error writing bgzf")
    }
    if err := gz.Close(); err != nil {
        t.Fatal("error closing bgzf")
    }

    b := buf.Bytes()[:16]
    expected := []byte(MagicBlock[:16])
    if !bytes.Equal(b, expected) {
        t.Fatalf("bgzf: incorrect header.\n got    %v\n wanted %v", b, expected)
    }
}
```
The output is:
```
--- FAIL: TestHeader (0.00s)
	bgzf_test.go:87: bgzf: incorrect header.
		 got    [31 139 8 20 0 0 0 0 0 0 11 0 66 67 2 0]
		 wanted [31 139 8 4 0 0 0 0 0 255 6 0 66 67 2 0]
FAIL
```

This is the line failing htslib checks: https://github.com/samtools/htslib/blob/6d927dfa3192492f3015b1bc9a0026f585e93acc/bgzf.c#L1494

kortschak:
I'd rather not conform to an expectation that is not specified.

kortschak:
@brentp Note that in the issue above, John invites a PR for the `bgzf_is_bgzf` behaviour change. I am not interested in doing that since I don't use tabix or, if at all possible, htslib and in the past PRs have languished there. You might want to push that forward if you need the behaviour to change.

kortschak:
I think this is not true now. John has merged a change to htslib that fixes this problem if our documentation is read.

brentp:
For bam, this means that we have to send a `bgzf.Writer` with `OS = 0xff`  to `bam.NewWriter` even with samtools 1.4

kortschak:
Why? The check in `bgzf_is_bgzf` does not require that OS be unknown.

I agree that the compress/gzip behaviour is unfortunate (actually that RFC1952's choice of OS unknown value is unfortunate), but this doesn't actually stop things working.

PR #58, opened 2017-01-05
bgzf: describe issues with tabix header checks

kortschak:
@brentp Please take a look.

Updates #57.

### DIFF
--- a/bgzf/bgzf_test.go
+++ b/bgzf/bgzf_test.go
@@ -28,6 +28,8 @@ import (
 )
 
 var (
+	go1_8 bool
+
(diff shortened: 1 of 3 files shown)


bgzf/bgzf_test.go
```diff
@@ -28,6 +28,8 @@ import (
 )
 
 var (
+	go1_8 bool
```
brentp: This is set by the runtime?

kortschak:
I'm going to leave this for a while until I find out what it going to happen with the htslib issues.

PR #61, opened 2017-01-14
bgzf: update spec compliance warning

kortschak:
@brentp Please take a look.

See https://github.com/samtools/htslib/commit/ec1d68e23ce75c7040ec895dfc0fffb1a2acb22c for rationale.


kortschak:
Note that this only applies to tabix built after that commit.

PR #62, opened 2017-01-22
handle case where b has fewer tags than a.

brentp:
without this, I see a panic on some bams. because it's looping over a and indexing into b without checking the length of b.


sam/header.go
```diff
@@ -111,6 +111,10 @@ type tagPair struct {
 	value string
 }
 
+func (t tagPair) String() string {
```
kortschak: Please remove this; `fmt.Stringer` gets in the way of debugging prints with `%+v`.

brentp:
added test-case, removed String() and force pushed.

brentp:
bump

PR #64, opened 2017-01-24
paper: extend summary to include citations

kortschak:
Addresses some of https://github.com/openjournals/joss-reviews/issues/157#issuecomment-274406225

@brentp Please take a look.

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -24,9 +24,14 @@ date: 6 January 2017
 
 # Summary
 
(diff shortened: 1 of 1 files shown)


kortschak:
@sb10 Please take a look. Addresses your comment https://github.com/biogo/hts/issues/66#issue-202845930

>There are also no DOI references. There is, for example, a published paper on the SAM spec out there.

PR #67, opened 2017-01-25
paper/examples/flagstat: use heavier benchmark and match logic

kortschak:
@brentp Please take a look.

This is a harsher benchmark and one that is used across implementations. We come out worse, but it is more realistic a test of use (the performance drop is likely due to the fact that GC becomes more of an issue for a long run). The change also makes the Go program a workalike to make comparison more convincing and simplifies some of the logic - we are still doing work that is unnecessary, but fixing that makes it more complicated.

### DIFF
--- a/paper/examples/flagstat/flagstat.go
+++ b/paper/examples/flagstat/flagstat.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// This package tabulates statistics on a bam file from the sam flag.
+// This program tabulates statistics on a bam file from the sam flag.
 // It replicates functionality in samtools flagstat.
 package main
 
@@ -13,6 +13,7 @@ import (
 	"os"
 
 	"github.com/biogo/hts/bam"
+	"github.com/biogo/hts/bgzf"
 	"github.com/biogo/hts/sam"
 )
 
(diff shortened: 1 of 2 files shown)


brentp:
I'll have a closer look tomorrow. I'm surprised there's such a difference between samtools and this (and between my comparison and yours). I was using the samtools from bioconda which is built with an old glibc and gcc so as to be compatible across machines; I wonder if that slows it down on my machine.

I guess we should also indicate the version of go was used.

paper/examples/flagstat/flagstat.go
```diff
@@ -30,17 +31,26 @@ func main() {
...
 
 	// counts is indexed by [pass/fail][sam.Flag] where we have 12 possible sam Flags.
 	var counts [2][12]uint64
 	// track mates on different chromosomes.
-	var mates [2]struct{ low, high uint64 }
+	var mates [2]struct{ allMapQ, highMapQ uint64 }
+	var good, singletons, paired [2]uint64
 	var qc int
 	for {
 		read, err := b.Read()
```
mnsmar: @brentp @kortschak I don't have a clean machine to check this properly, but I see roughly 10% better performance if `read` is declared outside the for loop to reduce allocations.

kortschak: Yeah, it's unlikely that this is causing that. The value doesn't escape to the heap, so it's not an expensive allocation.

kortschak:
I don't think that needs a note. What we want here is best opportunity for both candidates. Otherwise I would also have to note that if samtools is built without optimisations it also performs worse. What I did here was best time out of three runs for each tool.

PR #71, opened 2017-02-04
paper: properly mark up references

kortschak:
@brentp @arfon Please take a look.

Addresses https://github.com/openjournals/joss-reviews/issues/168#issuecomment-277406661


arfon:
Looks good but don't forget to reference the bibtex file in the paper metadata e.g.:

```
bibliography: paper.bib
```

PR #72, opened 2017-03-23
intial CRAM proposal

brentp:


proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+The `hts/cram` Reader should match the `hts/bam` API as closely as is reasonable.
+
+```Go
+
+import "biogo/hts/cram"
+
+var cr *cram.CRAM 
+var err error
+
+cr, err = cram.NewReader(rdr, io.Reader, rd int, reference *sam.Reference)
```
kortschak: `r io.Reader` (no comma and single letter).

proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+
+## High-Level API
+
+The `hts/cram` Reader should match the `hts/bam` API as closely as is reasonable.
+
+```Go
+
+import "biogo/hts/cram"
+
+var cr *cram.CRAM
```
kortschak: s/cram.CRAM/cram.Reader/

proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+
+```Go
+
+import "biogo/hts/cram"
+
+var cr *cram.CRAM 
+var err error
+
+cr, err = cram.NewReader(rdr, io.Reader, rd int, reference *sam.Reference)
+cr.Omit(bam.AllVariableLengthData)
```
kortschak: If we want finer control, this might not want to be a bam package constant, but rather in cram.

proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+
+var hdr *cram.Header = cr.Header()
+
+var rec *sam.Record
+rec, err = cr.Read()
+```
+
+The difference from the `hts/bam` API is the requirement of the `reference` argument to the constructor.
```
kortschak: Do we need to do this? Can you explain why cram needs the reference to be passed in?

proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+If the `sam.Reference` must be passed to the `Sequence()` method, then the `cram.NewReader` function would not
+need that value.
+
+We will need to understand how this would work with the regional query method in `hts/bam` that gets a slice of
+`[]bgzf.Chunk` to be sent to a `bam.Iterator`.
+
+the cram.Header will be identical to extend `sam.Header`:
+
+```Go
+type Header struct {
```
kortschak: What additional methods/fields are needed here?

proposal/cram.md
```diff
@@ -0,0 +1,95 @@
...
+### itf8
+
+`itf8` is a central data-type in `CRAM`.
+
+we can follow the implementation [here](https://github.com/samtools/htslib/blob/70622cffc711b0e501c431bc221a6abdbdf3a6bd/cram/cram_io.h#L111)
+
+with a signature like:
+
+```Go
+func readITF8(r io.reader) (int64, error)
```
kortschak: Consider a signature like [`utf8.DecodeRune`](https://golang.org/pkg/unicode/utf8/#DecodeRune) rather than something that takes an `io.Reader`. This actually matches the C implementation more closely.

proposal/cram.md
brentp: The spec isn't clear, though there's some words on it here: http://www.htslib.org/benchmarks/CRAM.html

and also `samtools view` on a cram requires the reference to be specified. I assumed for reads with a Cigar like `151M` it's not storing any bases. But haven't verified.

proposal/cram.md
brentp: I removed this. AFAICT it should be identical.

proposal/cram.md
brentp: I stubbed that out a tiny bit a while ago. the problem is that the `[]byte` input is not a fixed size so there's a lot of accounting of how to discard the read part of the `[]byte` and then combine the remainder with the rest of the data for the next read. Seemed simpler just to let it advance the reader as needed

brentp:
how do you mean "like sambamba"? system calls?

proposal/cram.md
kortschak: Have a look at the example for `utf8.DecodeRune`.

kortschak:
From what Artem wrote, sambamba calls into htslib. I'd rather not do this, but it's worth discussing and we might have to if we really want this soon.

kortschak:
I have written a reasonable proportion of the trivial plumbing for CRAM handling. This is in the [cram branch](https://github.com/biogo/hts/tree/cram). Please use it to play with to get an idea of things that will work and things that will be needed (for example, I don't think the concurrency setting will be needed in NewReader).

I think that in the long run it would be good to expose both a simple API similar to what you have above, but also an API that more closely reflects the CRAM structure. I'm writing this with that view in mind.

brentp:
sweet!

what's the API on block for dispatching blockData through the appropriate decompressor? are those defined on the cram.Reader or created for each block?

why no concurrency on the Reader?

is there a particular part I can contribute?

kortschak:
>what's the API on block for dispatching blockData through the appropriate decompressor? are those defined on the cram.Reader or created for each block?

TBD. At the moment I'm just building up the plumbing to see how to read through the data. The `TestRead` test function shows how this works. This is exploratory since the spec is minimal - I've had to break lots of things to figure out what some parts of the spec actually mean. Some of it I'm still not sure about.

>why no concurrency on the Reader?

At this stage it's unclear how it will fit in.

>is there a particular part I can contribute?

Feel free to just have a play. So far, I have low level handling of containers, blocks and slices sorted out. Bitstreams are unclear to me because in the core data block type I don't see how the length of each cram record is encoded.

Other things are the interaction with the codecs. At the moment, this is done ad hoc in `TestRead`. It should be a method on `block` that gives back a value depending on the `typ`. Exactly how to do this I'm not sure; I guess it could return an `interface{}`, but I'm not particularly keen on that.

PR #86, opened 2017-04-09
sam: fix CIGAR sequence length validation

brentp:


kortschak:
I'm not convinced this is the right fix, but if a zero length `Cigar` *is* valid for all sequence lengths then that should be documented and the early return should be done:

```
if len(c) == 0 {
    return true
}
```

However, an alternative is to check in [`Record.UnmarshalSAM`](https://github.com/biogo/hts/blob/master/sam/record.go#L258).

```
	if !bytes.Equal(f[9], []byte{'*'}) {
		r.Seq = NewSeq(f[9])
		if len(r.Cigar) != 0 && !r.Cigar.IsValid(r.Seq.Length) {
			return errors.New("sam: sequence/CIGAR length mismatch")
		}
	}
```

This better matches the semantics in the spec, that the information is not available so don't check, rather than an absent value means everything is valid.

(commit message should be "sam: fix CIGAR sequence length validation" - pref with an issue)

brentp:
done.

sam/record.go
```diff
@@ -255,7 +255,7 @@ func (r *Record) UnmarshalSAM(h *Header, b []byte) error {
 	}
 	if !bytes.Equal(f[9], []byte{'*'}) {
 		r.Seq = NewSeq(f[9])
-		if !r.Cigar.IsValid(r.Seq.Length) {
+		if !(len(r.Cigar) == 0 || r.Cigar.IsValid(r.Seq.Length)) {
```
kortschak: `len(r.Cigar) != 0 && !r.Cigar.IsValid(r.Seq.Length)`.

brentp: why is that better? I prefer the way I wrote it?

kortschak: When I read the negation distributed over the conjunction I need to understand both parts of the conjunction and then negate. With the disjunction I can understand as I read.

brentp: noted

brentp: and changed

PR #91, opened 2017-04-11
bam,sam: add merging functions

### LINKED ISSUE #77: bam: add Merge function
This package is absolutely a great work of Go for NGS data processing. I'm loving to use it. 
Is there an API for sorting bam files, similar to "samtools sort"?

kortschak:
@brentp @dongweigogo Please take a look.

Fixes #77.

### DIFF
--- a/bam/merger.go
+++ b/bam/merger.go
@@ -1,0 +1,187 @@
+// Copyright ©2017 The bíogo Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bam
+
+import (
+	"container/heap"
+	"errors"
+	"io"
+
+	"github.com/biogo/hts/sam"
+)
+
+// Merger implements merging BAM data with a defined sort
+// order. It can be used for sorting, concatenating and
+// deduplicating BAM data.
+type Merger struct {
+	h *sam.Header
+	// refLinks is the set of mappings from a RefID in
+	// a src Header to a Reference in the dst Header.
+	refLinks [][]*sam.Reference
+
+	less    func(a, b *sam.Record) bool
+	readers []*reader
+}
+
+type reader struct {
+	id int
+	r  *Reader
+
+	head *sam.Record
+	err  error
+}
+
+// NewMerger returns a Merger that reads from the source
+// Readers.
+//
+// The sort order of the stream merge is defined by the sort
+// order field of the src Reader headers and the provided less
+// function. The header sort order fields must agree.
+//
+// Sort order is determined using the following rules:
+//  - for sam.QueryName the LessByName sam.Record method is used.
+//  - for sam.Coordinate the LessByCoordinate sam.Record method is used.
+//  - for sam.Unsorted the reader streams are concatenated.
+//  - for sam.Unknown the provided less function is used - if nil
+//    this is the same as sam.Unsorted.
+// For all sort orders other than sam.Unknown, the less parameter
+// is ignored.
+// The src Readers should be closed individually after use to avoid
+// leaking resources.
+func NewMerger(less func(a, b *sam.Record) bool, src ...*Reader) (*Merger, error) {
+	if len(src) == 0 {
+		return nil, io.EOF
+	}
+
+	m := &Merger{readers: make([]*reader, len(src))}
+
+	headers := make([]*sam.Header, len(src))
+	so := src[0].Header().SortOrder
+	for i, r := range src {
+		h := r.Header()
+		if h.SortOrder != so {
+			return nil, errors.New("bam: sort order mismatch")
+		}
+		headers[i] = h
+	}
+	var err error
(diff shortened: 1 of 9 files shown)


bam/merger.go
```diff
@@ -0,0 +1,187 @@
...
+// NewMerger returns a Merger that reads from the source
+// Readers.
+//
+// The sort order of the stream merge is defined by the sort
+// order field of the src Reader headers and the provided less
+// function. The header sort order fields must agree.
+//
+// Sort order is determined using the following rules:
+//  - for sam.QueryName the LessByName sam.Record method is used.
+//  - for sam.Coordinate the LessByPostion sam.Record method is used.
```
brentp: LessByCoordinate

bam/merger.go
```diff
@@ -0,0 +1,187 @@
...
+//  - for sam.Unsorted the reader streams are concatenated.
+//  - for sam.Unknown the provided less function is used - if nil
+//    this is the same as sam.Unsorted.
+// For all sort orders other than sam.Unknown, the less parameter
```
brentp: I would put this above the previous block so that users can skip that section if they know they have unknown.

bam/merger.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// NewMerger returns a Merger that reads from the source
+// Readers.
+//
+// The sort order of the stream merge is defined by the sort
+// order field of the src Reader headers and the provided less
+// function. The header sort order fields must agree.
+//
+// Sort order is determined using the following rules:
```
brentp: The sort function is determined ...

sam/record.go
```diff
@@ -173,6 +173,25 @@ func (r *Record) Strand() int8 {
...
 
+// LessByName returns true if the receiver sorts by record name before other.
+func (r *Record) LessByName(other *Record) bool {
+	return r.Name < other.Name
+}
+
+// LessByCoordinate returns true if the receiver sorts by coordinate before other
+// according to the SAM specification.
+func (r *Record) LessByCoordinate(other *Record) bool {
+	rRefName := r.Ref.Name()
```
brentp: can you use `Ref.ID()` for these instead?

sam/header.go
```diff
@@ -449,6 +540,21 @@ func (bh *Header) AddReadGroup(rg *ReadGroup) error {
 	return nil
 }
 
+// RemoveReadGroup removes rg from the Header and makes it
+// available to add to another Header.
+func (bh *Header) RemoveReadGroup(rg *ReadGroup) error {
```
brentp: no tests for this?

brentp: I guess it's exercised by the merge tests...

bam/merger.go
kortschak: It's not necessarily a sort function.

bam/merger.go
kortschak: Done.

bam/merger.go
kortschak: It really just reiterates the last bullet.

sam/record.go
kortschak: The SAM spec says that coordinate sort is based on the name of the reference. The reason for this is that that value is stable while the ref id depends on the order of references in the header.

sam/header.go
kortschak: It's not tested by those. I should add tests for these.

kortschak:
>One of the things I like about biogo/hts is that it's pretty minimal, this seems like a nice example of what you can do with the library, but it adds quite a bit of code. (I'm not sure how much of it is specific to this application and how much is really needed).

Much of what is added here is stuff which really should already have existed for manipulating headers. The majority of the rest is tests. Less than 200 lines is new code for merging files. The benefits from this addition though is that arbitrary sorting is possible and with that the capacity to perform single end and paired end read deduplication (in the correct manner, as opposed to the way samtools does it). I think the cost benefit is well in favour of the addition.

>People are asking for examples. This seems like something that could be an application example.

There is an example for sorting that should be just copy-pastable.

sam/header.go
kortschak: Done.

kortschak:
PTAL

bam/merger.go
```diff
@@ -0,0 +1,187 @@
...
+	if len(m.readers) == 0 {
+		return nil, io.EOF
+	}
+	if m.less == nil {
+		return m.cat()
+	}
+	return m.nextBySortOrder()
+}
+
+func (m *Merger) cat() (rec *sam.Record, err error) {
```
brentp: is there no way to have a special sort order for this case so that you can drop `cat` and just use `nextBySortOrder` with a less function that just orders by reader index?

something like:

```
func (m *bySortOrderAndID) Less(i, j int) bool {
    if m.less == nil {
        return i <= j
    }
    ...
}
```

kortschak: This adds a not insignificant load to something that needs almost no work for correctness. The `cat` helper method is only a few lines and trivially simple.

PR #95, opened 2017-04-15
bam: fix merger for short bams

### LINKED ISSUE #94: The merge process panics when the bam is small
For a small size bam file in which the records number is less than the chunk size, the process panics. As the godoc example:
I first found one of the reasons is that the records length n is not updated when the first chunk is not filled up.

    for {
        for it.Next() {
            recs = append(recs, it.Record())
            if len(recs) == cap(recs) {
                r, err := writeChunk(dir, h, recs)
                if err != nil {
                    log.Panic(err)
                }
                t = append(t, r)
                n, recs = len(recs), recs[:0]     //when the records is less than the chunk size, n will not be updated here
            }
        }
        err = it.Error()
        if n == 0 || err != nil {    //break because n is not updated, and thus the reader slices t is still empty without any reader appended.
            break
        }
        if len(recs) != 0 {
            r, err := writeChunk(dir, h, recs)
            if err != nil {
                log.Panic(err)
            }
            t = append(t, r)
        }
    }

I think that can be fixed by just simply changing it to "n++" for every record in the iterator. When I did this, the reader slice t is updated with one reader as expected. However it panics once again when it goes to:

    m, err := bam.NewMerger(nil, t...)
    if err != nil {
        log.Panicf("failed to created merger: %v", err)
    }
    sorted := sam.NewIterator(m)
    for sorted.Next() {                              //here
        // Operate on coordinate sorted stream.
        fn(sorted.Record())
    }

I don't know what happens here. I'm guessing if it's about something in the merger constructor when there is only one underlying reader.

BTW, that does not happen on big bam files in which case there are more than one sub-reader.

kortschak:
@brentp @dongweigogo Please take a look.

Fixes #94.

### DIFF
--- a/bam/merger.go
+++ b/bam/merger.go
@@ -154,7 +154,7 @@ func (m *Merger) nextBySortOrder() (rec *sam.Record, err error) {
 }
 
 func (m *Merger) reassignReference(id int, rec *sam.Record) {
-	if rec.Ref == nil {
+	if rec.Ref == nil || m.refLinks == nil {
 		return
 	}
 	rec.Ref = m.refLinks[id][rec.RefID()]
--- a/bam/merger_example_test.go
+++ b/bam/merger_example_test.go
@@ -75,16 +75,17 @@ func ExampleMerger_sortByCoordinate() {
 				n, recs = len(recs), recs[:0]
 			}
 		}
-		err = it.Error()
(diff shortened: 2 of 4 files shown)


dongweigogo:
Great, it seems to work well.

PR #100, opened 2017-07-05
sam: conform more closely to ISO8601

kortschak:
@Defgenx Please try this change.


Defgenx:
Just tested it and it works with my bam file.

Thanks a lot @kortschak .

PR #103, opened 2017-08-09
sam: reflect the behaviour of other tools

kortschak:
@brentp Please take a look.

This change upsets me, but we should fit in with the behaviour of other tools in the ecosystem, I guess.

### DIFF
--- a/sam/record.go
+++ b/sam/record.go
@@ -131,10 +131,6 @@ func (r *Record) Bin() int {
 	if r.Flags&(Unmapped|MateUnmapped) == Unmapped|MateUnmapped {
 		return 4680 // reg2bin(-1, 0)
 	}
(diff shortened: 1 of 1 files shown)


sam/record.go
```diff
@@ -131,10 +131,6 @@ func (r *Record) Bin() int {
 	if r.Flags&(Unmapped|MateUnmapped) == Unmapped|MateUnmapped {
 		return 4680 // reg2bin(-1, 0)
 	}
-	end := r.End()
-	if !internal.IsValidIndexPos(r.Pos) || !internal.IsValidIndexPos(end) {
-		return -1
-	}
 	return int(internal.BinFor(r.Pos, r.End()))
```
brentp: re-use `end` here.

PR #105, opened 2017-08-19
sam: make zero length CIGAR or unmapped reads have unit length

### LINKED ISSUE #104: Zero length bam records have invalid bin values
I encountered a problem with biogo's bam output for records that have a zero length alignment, but a non-zero position and reference.  This case can happen when the data has a read pair where one read is mapped, and the mate is unmapped.  The unmapped mate will have the same position and reference as the mapped mate.  In this case, the unmapped mate will have a zero length alignment.

When I output these records to a bam file using biogo's bam writer, and validate the output bam using "picard ValidateSamFile", I get errors like this:
```
ERROR: Record 105026, Read name FOOBAR, bin field of BAM record does not equal value computed based on alignment start and end, and length of sequence to which read is aligned
```
"sambamba_v0.6.6 index -c" reports 
```
sambamba-index: Bin in read with name 'FOOBAR' is set incorrectly (810 instead of expected 6485)
```

I traced this problem back to a difference between picard and biogo's handling of bin calculation.  When picard (and I assume sambamba) encounters a read with 0 length alignment, it adds 1 to the end position before calling reg2bin(), thus considering the alignment to have length 1.  I'm guessing this is a special case that was not mentioned in the SAM specification.  See https://github.com/samtools/htsjdk/blob/c20e0edd361189c3f2bc3718b018dac2ec90530b/src/main/java/htsjdk/samtools/SAMRecord.java#L1554

Here's a diff I made to record.go that addresses the problem in my use case:
```
@@ -132,10 +132,18 @@ func (r *Record) Bin() int {
                return 4680 // reg2bin(-1, 0)
        }
        end := r.End()
+
+       // If the alignment length is zero (for example, if the read is
+       // unmapped), then increment end by 1 and treat the read as length
+       // 1 for binning purposes.
+       if end == r.Pos {
+               end++
+       }
+
        if !internal.IsValidIndexPos(r.Pos) || !internal.IsValidIndexPos(end) {
                return -1
        }
-       return int(internal.BinFor(r.Pos, r.End()))
+       return int(internal.BinFor(r.Pos, end))
 }
```

kortschak:
@brentp @yipal Please take a look.

Fixes #104.

### DIFF
--- a/sam/record.go
+++ b/sam/record.go
@@ -150,6 +150,9 @@ func max(a, b int) int {
 // The position returned by End is not valid if r.Cigar.IsValid(r.Seq.Length)
 // is false.
 func (r *Record) End() int {
+	if r.Flags&Unmapped != 0 || len(r.Cigar) == 0 {
+		return r.Pos + 1
+	}
 	pos := r.Pos
 	end := pos
 	for _, co := range r.Cigar {


brentp:
I see this matches `bam_endpos`. Is that the goal here?
otherwise, it seems like modifying Bin as proposed in #104 is a more direct fix for the bin calculation. 

The relevant part of the spec quoted in 104 is mentioned in the context of the bin calc, only...

kortschak:
>I see this matches `bam_endpos`. Is that the goal here?

Yes.

>The relevant part of the spec quoted in 104 is mentioned in the context of the bin calc, only...

The smaller the number of special-casings the better. For better or worse the spec for SAM/BAM lies in the implementations and the text of the spec generally trails those. There is no other explicit explanation of how to determine the length of an alignment except in the code of the implementations (there are discussions on samtools-devel as well), maybe there could be an issue filed against hts-specs to make this less ambiguous. I will not be the person to do that.

In reality, the impact of the change here is minimal for real world uses.

brentp:
LGTM, in that case.

PR #108, opened 2017-10-05
Added fuzz testing for bgzf and bam readers.

yasushi-saito:
Fixed a few bugs found by them.

### DIFF
--- a/bam/reader.go
+++ b/bam/reader.go
@@ -33,6 +33,9 @@ type Reader struct {
 	buf [4]byte
 }
 
+// Max possible bytesize of one record.
+const maxBAMRecordSize = 0xfffffff
+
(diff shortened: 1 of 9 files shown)


kortschak:
I think this is worth adding, but the structure of the PR cannot be acceepted; there is too much happening to clearly review and if a small change here causes breaking brehaviour bisection becomes very difficult.

Can you:

1. Send a PR adding the fuzzer code and yourself to A+C.
2. File issues for the bugs you have found running the fuzzer.
3. Send PR for each bug individually.

I understand that this takes more work in the short term, but makes it much easier to ensure that good changes are being made and can be properly reasoned about in the future.

Thanks.

kortschak:
I'm closing this since the fuzzed crashers have been reproduced and fixed elsewhere.

PR #109, opened 2017-10-12
bgzf: fix block size handling errors

kortschak:
Found by go-fuzz, hinted at by #108.

I re-ran the bgzf go-fuzz code at the issue above and can more simply fix the errors by handling the zero and negative needed limited read cases.

I am not entirely sure of the zero-size case. I suspect that the `io.EOF` should be revised to a `nil` error after it has been intercepted after `d.readMember` returns in `decompressor.nextBlockAt`, but I would prefer this failure to the panic that exists, though please have a think about that.

I think the fuzzing has saturated, but I'll leave it running for a while.

@brentp Please take a look.

### DIFF
--- a/bgzf/bgzf.go
+++ b/bgzf/bgzf.go
@@ -45,6 +45,7 @@ func init() {
 
 var (
 	ErrClosed            = errors.New("bgzf: use of closed writer")
+	ErrCorrupt           = errors.New("bgzf: corrupt block")
 	ErrBlockOverflow     = errors.New("bgzf: block overflow")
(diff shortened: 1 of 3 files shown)


kortschak:
I have tried using a private error and nilling it out after it has been detected and this leads to a deadlock, but I'm still suspicious. I think I should check whether a magic block followed by a non-empty block returns the contents of the non-empty block. This is simple to check and I'll do it tomorrow.

kortschak:
I have confirmed that this is safe (at least for the magic block). PTAL

PR #110, opened 2017-10-13
bam,sam: fix zero and negativish value handling

kortschak:
@brentp Please take a look.


kortschak:
I am leaving the fuzzer running on the fixed code.

kortschak:
Found a couple more that can be protected against. There are more, but they appear to be just massive over allocation that only cause a problem in a low memory environment. I'll make a new issue for those.

PTAL

bam/reader.go
```diff
@@ -355,9 +355,17 @@ func parseAux(aux []byte) []sam.Aux {
...
 
 func (b *buffer) bytes(n int) []byte {
+	if b.err != nil {
+		return nil
+	}
+	if b.len() < n {
+		b.err = io.ErrUnexpectedEOF
+		return nil
+	}
 	s := b.off
```
brentp: wouldn't it be better to check that `b.len() < n + b.off` ?

kortschak: `b.len()` is the number of remaining bytes in the buffer (see [here](https://github.com/biogo/hts/blob/d0dadd4feeafd9428b7186d376fa658ab1ac075d/bam/reader.go#L374)).

brentp: ah. right.

kortschak:
I'm going to merge this though it does not fix all crashers. The remaining crashers seem to be only occurring when go-fuzz is running with multiple slaves, thus allowing memory exhaustion by concerted effort. I'm still not happy with the fact that small inputs can cause this, but it's reasonable I guess. I'll follow those at #111, and I'm continuing to run go-fuzz on the current state with a single slave to flush out any other cases here.

PR #112, opened 2017-10-18
bam: make bad auxtags and sequence lengths safe

kortschak:
@brentp Please take a look.


kortschak:
This has fixed most of the crashers that aren't due to cooperative memory exhaustion. I have some more over night, but I'll get to those later.

PR #114, opened 2017-10-23
bam: fix yet more bounds checks

kortschak:
This is getting tedious.

@brentp Please take a look.


kortschak:
I'm pretty sure that `i+j < 0` costs about as much as `i < j` when it's compiled since you either have and add and test against zero, or a cmp (subtract without return) and a test against zero.

Yes.

`i+j < 0`:
```
	0x0034 00052 (cmp.go:10)	MOVQ	"".i+48(SP), AX
	0x0039 00057 (cmp.go:11)	ADDQ	(SP), AX
	0x003d 00061 (cmp.go:11)	TESTQ	AX, AX
	0x0040 00064 (cmp.go:11)	JLT	76
```

`i < j`:
```
	0x0034 00052 (cmp.go:10)	MOVQ	(SP), AX
	0x0038 00056 (cmp.go:10)	MOVQ	"".i+48(SP), CX
	0x003d 00061 (cmp.go:11)	CMPQ	CX, AX
	0x0040 00064 (cmp.go:11)	JLT	76
```

brentp:
interesting. I was thinking after I posted that it was probably transformed by the SSA.

kortschak:
No, that's just how less than is implemented in a processor - even the old 6502 works that way; `i < j` is expressed as `i-j < 0`, throwing away the result, but setting the relevant borrow and zero flags.

PR #117, opened 2017-12-13
csi: don't break too early when adding chunks

### LINKED ISSUE #116: csi: missing chunks from CSI index
I've updated [vcfanno](https://github.com/brentp/vcfanno) to use CSI when available. I did some exhaustive testing and found that in very rare cases, there are missed annotations.
I'm sure this is due to logic in the `Chunks()` method. I have not been able to make a test-case, but I will make a PR that fixes this shortly.

I think it may be related to the logic change that you did for tabix (internal/index) here https://github.com/biogo/hts/pull/20/files

brentp:
closes #116

### DIFF
--- a/csi/csi.go
+++ b/csi/csi.go
@@ -220,6 +220,8 @@ func (i *Index) Chunks(rid int, beg, end int) []bgzf.Chunk {
 			for _, chunk := range ref.bins[c].chunks {
 				if vOffset(chunk.End) > left {
 					chunks = append(chunks, chunk)
+				}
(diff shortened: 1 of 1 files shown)


kortschak:
Can we add the test case please?

brentp:
I didn't find how to generate a test-case. I found this doing a whole-genome functional test. 

I'll try to carve out some time next week to have another look.

brentp:
I found a case where this change is also incorrect.

PR #118, opened 2017-12-20
csi,internal: fix stats dummy number handling

### LINKED ISSUE #115: csi: malformed dummy bin header
On csi files created by htslib (with bcftools index or with tabix), I see this error.

I think there must be something wrong with the check: `if bins[i].bin == statsDummyBin` in csi_read.go that causes this (given enough data) to trigger when n != 2. I can't find in the spec or htslib code how you came up with that. Is there any other (currently missing) constraint for this check?

For most files, I do not have this problem, but for 1 large file I do.

This is 1 of 2 issues related to CSI that I've found. Opening the other issue presently.

kortschak:
Please take a look.

Fixes #115.


kortschak:
Note that the code quoted [here](https://sourceforge.net/p/samtools/mailman/message/36157993/):

```
/* calculate maximum bin number -- valid bin numbers range within [0,bin_limit) */
int bin_limit(int min_shift, int depth)
{
    return ((1 << (depth+1)*3) - 1) / 7;
}
```

gives 299593 when the [SAM spec-specified depth of 6](https://github.com/samtools/hts-specs/blob/7aa24be3d856b7af9eabf4533c227be5efac0f6f/SAMv1.tex#L1044-L1048) is used, instead of [SAM spec-specified value of 37449](https://github.com/samtools/hts-specs/blob/7aa24be3d856b7af9eabf4533c227be5efac0f6f/SAMv1.tex#L1027-L1029) which is given with a depth of 5.

kortschak:
Thanks

PR #119, opened 2018-03-12
Add GetAllOffsets() to retrieve block offsets from a BAM index

yipal:
It is possible to use the offsets from the .bai index file to find
block start positions in the corresponding .bam file.  The block start
positions, are useful for dividing a bam file into approximately equal
sized pieces for subsequent processing.

Also fix a compile error in bam/merger_example_test.go.

### DIFF
--- a/bam/index.go
+++ b/bam/index.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
+	"sort"
 
 	"github.com/biogo/hts/bgzf"
 	"github.com/biogo/hts/bgzf/index"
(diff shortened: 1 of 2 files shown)


yipal:
Hi Dan, I have a small feature I use in biogo that exposes the block offsets in the .bai file to users.  I thought it would be helpful to other biogo users if it were part of the library.  It would also make it easier for me to opensource some code I have that uses biogo, since my code could use the official biogo library without any modifications.  Here's the change.

kortschak:
This seems like a complicated way to do something that can be done in a few lines using the current API.

```
func allChunksFor(r *sam.Reference, bai *bam.Index) []bgzf.Chunk {
    return bai.Chunks(r, 0, r.Len())
}
```
with the map handling dealt with by the user

or more aligned to how you have it,

```
func allChunksFor(bai *bam.Index) map[int][]bgzf.Chunk {
    cm := make(map[int][]bgzf.Chunk)
    for _, r := 0; r < bai.NumRefs(); r++ {
        cm[i] = bai.Chunks(r, 0, 1<<30)
    }
    return cm
}
```

Neither are tested, but both should work.

On a style note, `Get` is not a prefix that is used for Go accessors.

yipal:
Thanks Dan!  I tried this:
`bai.Chunks(r, 0, r.Len())`

And found that Chunks() merges adjacent chunks using Adjacent()

```
func (i *Index) Chunks(r *sam.Reference, beg, end int) ([]bgzf.Chunk, error) {
    chunks, err := i.idx.Chunks(r.ID(), beg, end)
    if err != nil {
        return nil, err
    }
    return index.Adjacent(chunks), nil
}

```

So that my code will see only a single Chunk, whereas I want to see all the chunks even if they are adjacent.  Is there another way where I can get all the Chunks?  If not, I can add a variation on Chunks() which accepts a MergeStrategy as a parameter, and then I can pass Identity to avoid the merging.

kortschak:
Can you explain the background need here? The architecture that you allude to seems interesting and I'd like to understand it better to see if it is something that should be supported.

yipal:
Sure, I have some large .bam files 100-500GB and I have code that divides a single .bam file into many smaller shards.  My code then reads and processes each shard using a worker pool so that I can use all the CPUs on my machine.  Ideally, the shards are roughly the same size to avoid having some shards that take much longer to process than the other shards (larger shards end up running by themselves at the end of the run, without using all available CPUs).

So what I have been doing is reading all the Chunks from the .bai index, and dividing the .bam file into shards using the Offset.File as a way to know how many bytes each shard would have.  Once I have the file divided, each worker can seek to the part of the file using the Offsets I got from the index.

kortschak:
Is there any reason not to just send records on a worker queue to a the pool of workers? (What is the relative effort that a single read needs?) Alternatively hand the `bam.Reader` around between the worker pool go routines, passing it on when a worker has enough records to make the work worthwhile. I suspect that you are working against the `bgzf.Reader`'s read scheduling by seeking around with more than one reader (both in seek time and BGZF block decompression work duplication) - in many cases this will not be a problem, but the engineering here is brittly dependent on assumptions that do not always hold.

Have a think about that, and if you still think that it is a worthwhile thing to do, then a `MergeStrategy` field can be added to `bam.Index` which will be used if not nil, otherwise using `index.Adjacent`.

yipal:
We were not able to fully use all cores when dispatching records to workers from a central reader.  We didn't spend the time to collect concrete evidence, but we suspect that the limitation is that cross-cpu communication ends up being very expensive on machines with 32-64 cores.  We think passing each record between cores is too fine grained in our use case.

Also, since we are using SSD storage, we don't pay a big penalty for reading from more than one place in the file, so using 64 workers each reading from a different part of the file avoids cross communication cost between cores without paying high costs for seeking.

I don't think I understand your proposal for passing a bam.Reader around entirely, but that sounds like we'd need to pay cross-cpu coordination costs, and maybe buffering a bunch of bam records in memory for processing while the other cores get their chance to read.

For our uses, we believe that sharding at the very start, and letting each core do its thing separately gives us the best performance.  We can scan a 130GB bam file in under 15 minutes fully utilizing 40/40 cores.

I'll prepare a patch for a MergeStrategy parameter.

kortschak:
>I don't think I understand your proposal for passing a bam.Reader around entirely, but that sounds like we'd need to pay cross-cpu coordination costs, and maybe buffering a bunch of bam records in memory for processing while the other cores get their chance to read.

Buffering would be required in an adequately large work slice for each worker, though this is just a pointer assignment for each record into a one-time allocated slice per worker, and there would be a `chan *bam.Reader` , but coordination costs are swamped by flate costs.

>For our uses, we believe that sharding at the very start, and letting each core do its thing separately gives us the best performance. We can scan a 130GB bam file in under 15 minutes fully utilizing 40/40 cores.

This really depends on what the relative amount of time for doing work on a record is compared to getting a record from the BAM. You really should measure these for your system and work load.

kortschak:
Closing this, but I'd still be interested in seeing what the difference in performance is, or at least the relative cost of processing a single record in your work flow, compared to obtaining the record.

PR #122, opened 2018-03-22
Add methods to sam.Seq to get the N'th base without expanding.

yasushi-saito:
Before, to get the base at a specific position in a Seq, we needed to call
Expand, which was quite expensive. This change adds Seq.Base() to get the base
type directly.

Also add helper functions to convert between BAM's sequence encoding and
human-readable characters.


sam/record.go
```diff
@@ -451,6 +451,32 @@ var (
...
+	BaseT    SeqBase = 8
+	BaseW    SeqBase = 9
+	BaseY    SeqBase = 10
+	BaseH    SeqBase = 11
+	BaseK    SeqBase = 12
+	BaseD    SeqBase = 13
+	BaseB    SeqBase = 14
+	BaseN    SeqBase = 15
+
+	// NumSeqBaseTypes is number of possible SeqBase values.
```
kortschak: This is not necessary.

sam/record.go
```diff
@@ -490,3 +516,32 @@ func (ns Seq) Expand() []byte {
 
 	return s
 }
+
+// Base returns the pos'th base of the sequence.  "pos" must be smaller than
+// seq.Length.
+func (ns Seq) Base(pos int) SeqBase {
```
kortschak: I would prefer this be named `At` and return a `byte` which is actually the encoded letter, so just what your `BaseChar` does below.

The documentation here should state that the call will panic if `pos < 0 || ns.Len() <= pos`, and do this explicit check and panic with an informative message.

kortschak:
Thanks for this and apologies for the delay in responding.

PR #128, opened 2018-12-11
hts: Address unchecked-error warnings

yasushi-saito:
They were found by errcheck.  As for errWriter / binaryWriter, this change
merges the two types and changes write* methods not to return any error. Writing
to bytes.Buffer is guaranteed to be error free.

### DIFF
--- a/bam/writer.go
+++ b/bam/writer.go
@@ -89,8 +89,7 @@ func (bw *Writer) Write(r *sam.Record) error {
 		len(tags)
 
 	bw.buf.Reset()
-	wb := errWriter{w: &bw.buf}
-	bin := binaryWriter{w: &wb}
+	bin := binaryWriter{w: &bw.buf}
 
 	// Write record header data.
 	bin.writeInt32(int32(recLen))
@@ -107,21 +106,18 @@ func (bw *Writer) Write(r *sam.Record) error {
 	bin.writeInt32(int32(r.TempLen))
 
 	// Write variable length data.
-	wb.Write(append([]byte(r.Name), 0))
+	bw.buf.WriteString(r.Name)
+	bw.buf.WriteByte(0)
 	writeCigarOps(&bin, r.Cigar)
-	wb.Write(doublets(r.Seq.Seq).Bytes())
+	bw.buf.Write(doublets(r.Seq.Seq).Bytes())
 	if r.Qual != nil {
-		wb.Write(r.Qual)
+		bw.buf.Write(r.Qual)
 	} else {
 		for i := 0; i < r.Seq.Length; i++ {
-			wb.WriteByte(0xff)
+			bw.buf.WriteByte(0xff)
 		}
 	}
-	wb.Write(tags)
-	if wb.err != nil {
-		return wb.err
-	}
-
+	bw.buf.Write(tags)
 	_, err := bw.bg.Write(bw.buf.Bytes())
 	return err
 }
@@ -129,11 +125,7 @@ func (bw *Writer) Write(r *sam.Record) error {
 func writeCigarOps(bin *binaryWriter, co []sam.CigarOp) {
 	for _, o := range co {
 		bin.writeUint32(uint32(o))
-		if bin.w.err != nil {
-			return
-		}
 	}
-	return
 }
 
 // Close closes the writer.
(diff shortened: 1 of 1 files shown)


bam/reader.go
```diff
@@ -296,7 +296,9 @@ func (i *Iterator) Record() *sam.Record { return i.rec }
 
 // Close releases the underlying Reader.
 func (i *Iterator) Close() error {
-	i.r.SetChunk(nil)
+	if err := i.r.SetChunk(nil); err != nil {
```
kortschak: This isn't necessary. `SetChunk(nil)` always returns `nil`.

bam/writer.go
```diff
@@ -53,7 +53,10 @@ func NewWriterLevel(w io.Writer, h *sam.Header, level, wc int) (*Writer, error)
 	if err != nil {
 		return nil, err
 	}
-	bw.bg.Flush()
+	err = bw.bg.Flush()
```
kortschak: This is not necessary. `error`s in `bgzf.Writer` are sticky, so any failure is caught at the call to `Wait` that follows.

bam/writer.go
```diff
@@ -107,83 +109,72 @@ func (bw *Writer) Write(r *sam.Record) error {
...
+	if err := w.w.WriteByte(b); err != nil {
+		panic(err)
 	}
-	w.err = w.w.WriteByte(b)
-	return w.err
 }
 
 type binaryWriter struct {
-	w   *errWriter
+	w   *bytes.Buffer
```
kortschak:
```suggestion
	// No error handling on the buffer since it
	// is documented to always return nil err.
	*bytes.Buffer
```

Then all the writes are just `Write`/`WriteByte`/`WriteString` calls on the `bytes.Buffer` and the three methods above can be deleted.

bam/writer.go
```diff
@@ -107,83 +109,72 @@ func (bw *Writer) Write(r *sam.Record) error {
...
 // Close closes the writer.
 func (bw *Writer) Close() error {
 	return bw.bg.Close()
 }
 
-type errWriter struct {
-	w   *bytes.Buffer
-	err error
+func (w *binaryWriter) write(p []byte) {
+	if _, err := w.w.Write(p); err != nil {
```
kortschak: ~These are documented to not be necessary as you mention, so don't bother with this.~

bam/writer.go
```diff
@@ -107,83 +109,72 @@ func (bw *Writer) Write(r *sam.Record) error {
...
+		panic(err)
+	}
 }
 
-func (w *errWriter) Write(p []byte) (int, error) {
-	if w.err != nil {
-		return 0, w.err
+func (w *binaryWriter) writeString(p string) {
+	if _, err := w.w.WriteString(p); err != nil {
```
kortschak: ~Same here.~

bam/writer.go
```diff
@@ -107,83 +109,72 @@ func (bw *Writer) Write(r *sam.Record) error {
...
-	var n int
-	n, w.err = w.w.Write(p)
-	return n, w.err
 }
 
-func (w *errWriter) WriteByte(b byte) error {
-	if w.err != nil {
-		return w.err
+func (w *binaryWriter) writeByte(b byte) {
+	if err := w.w.WriteByte(b); err != nil {
```
kortschak: ~And here.~

bam/writer.go
```diff
@@ -107,83 +109,72 @@ func (bw *Writer) Write(r *sam.Record) error {
...
 type binaryWriter struct {
-	w   *errWriter
+	w   *bytes.Buffer
 	buf [4]byte
 }
 
 func (w *binaryWriter) writeUint8(v uint8) {
 	w.buf[0] = v
-	w.w.Write(w.buf[:1])
+	w.write(w.buf[:1])
```
kortschak: These four all become `w.Write`.

bam/reader.go
yasushi-saito: Ok. I prefer having a simple check just to be sure. But it's your call.

yasushi-saito:
I address you comments. Can you take another look?

kortschak:
Thanks. The code LGTM. Would you please change the commit message to 

```
bam: elide unnecessary error checks

The Write method is guaranteed to return a nil error, so these checks are not needed.
```

kortschak:
Thanks

PR #130, opened 2019-02-24
cram: add package for building cram support on

kortschak:
@brentp I think we should merge this as a basis for continuing work. It's not going to attract  work in its own branch.

I will add docs and have a read through over the next couple of days, so I won't need an approval until then.


kortschak:
PTAL @brentp 

Also @sb10, please feel free to have a look.

cram/cram.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+// readFrom populates a definition from the given io.Reader. If the magic
+// number of the file is not "CRAM" readFrom returns an error.
+func (d *definition) readFrom(r io.Reader) error {
+	err := binary.Read(r, binary.LittleEndian, d)
+	if err != nil {
+		return err
+	}
+	magic := reflect.TypeOf(*d).Field(0).Tag.Get("is")
```
brentp: this seems like a complicated way to get `[]byte("CRAM")`

cram/cram.go
```diff
@@ -0,0 +1,488 @@
...
+	case bzip2Method:
+		return ioutil.ReadAll(bzip2.NewReader(bytes.NewReader(b.blockData)))
+	case lzmaMethod:
+		lz, err := lzma.NewReader(bytes.NewReader(b.blockData))
+		if err != nil {
+			return nil, err
+		}
+		return ioutil.ReadAll(lz)
+	case ransMethod:
+		// Unimplemented.
```
brentp: this could panic or return error

cram/encoding/ltf8/ltf.go
```diff
@@ -0,0 +1,170 @@
...
+	5:  6,
+	6:  1,
+	9:  4,
+	10: 2,
+	11: 5,
+	14: 0,
+}
+
+// nlo returns the number of leading set bits in x.
+func nlo(x byte) int {
```
brentp: `nlo(x)` (and pop) can be replace by `bits.LeadingZeros8(^x)`

cram/encoding/itf8/itf.go
```diff
@@ -0,0 +1,112 @@
...
+	5:  6,
+	6:  1,
+	9:  4,
+	10: 2,
+	11: 5,
+	14: 0,
+}
+
+// nlo returns the number of leading set bits in x.
+func nlo(x byte) int {
```
brentp: repalce with `bits.LeadingZeros8(^x)`

cram/encoding/ltf8/ltf.go
brentp: I guess `bits` dependency would truncate support to more recent versions of go.

kortschak: I'm happy to drop that. "math/bits" didn't exist when I wrote this.

cram/cram.go
kortschak: At the time, it helped me keep the ownership details under control. If it's a performance problem we can revisit.

cram/encoding/itf8/itf.go
kortschak: Done

cram/encoding/ltf8/ltf.go
kortschak: Done

cram/cram.go
kortschak: Done.

PR #132, opened 2019-06-12
Fixed record length calculation in BAM writer when qualities are absent.

bsipos:
Fixed record length calculation in BAM writer when qualities are absent.


kortschak:
There is a simpler fix. I'll send this later.

PR #134, opened 2019-06-13
bam: fix qual encoding length for nil qualities

kortschak:
Fixes defect found in #132 more simply.

Please take a look.

/cc @bsipos


bsipos:
Indeed this is better.

kortschak:
Closes #132.

PR #135, opened 2019-06-13
sam: don't compare query length to template length in validation

### LINKED ISSUE #133: Possible issue with SAM record validation
The line below compares the length of the sequence stored in a sam.Record to record.Len(). However, Len() calculates the number of bases consumed on the reference and not on the query.

https://github.com/biogo/hts/blob/aafaa03b4ed961cf75d5f7ace0d70cf9fa921f64/sam/record.go#L99

kortschak:
@bsipos Please take a look.

Would you try this with your code? If you can provide a simple reproducer, I will include that in a test here.

Fixes #133.

### DIFF
--- a/sam/record.go
+++ b/sam/record.go
@@ -96,10 +96,18 @@ func IsValidRecord(r *Record) bool {
 	if len(r.Qual) != 0 && r.Seq.Length != len(r.Qual) {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


bsipos:
Works for me!

Any record with a soft clip for example would trigger the problem:

`q3      0       q2      4       7    1S13M2I18M *       0       0       TCAGGAAGGCGGTATCGACGACGGTGCCGAGGCA      *       PG:i:0  NM:i:0`

PR #136, opened 2019-06-13
 Added NewRecordFromAln for building SAM records from "raw" local alignments.

bsipos:
This function creates a new SAM record by parsing a "raw" local alignment such like this:

`CAGGAAGGCGGTATCGACGACGGTGCCGAGGCA`
`CAGGAAGGCGGTA--GACGACGGTGCCGAGGCA `

### DIFF
--- a/bam/writer.go
+++ b/bam/writer.go
@@ -81,11 +81,15 @@ func (bw *Writer) Write(r *sam.Record) error {
 		return errors.New("bam: sequence/quality length mismatch")
 	}
 	tags := buildAux(r.AuxFields)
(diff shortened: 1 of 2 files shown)


kortschak:
I need to think about this. I'm not sure it's in scope for the package.

@brentp Do you have any opinions on this?

brentp:
I can see the utility of this, but I'd rather keep biogo/hts minimal so it's easy find what's needed for most cases.

bsipos:
@kortschak @brentp Okay, I will just add it to my codebase then.

PR #148, opened 2020-01-24
fai: implement FAI reading and faidx use

kortschak:
This brings the fai package from biogo/biogo to here and implements faidx reading using mmap.

When this is merged, I will make the biogo/biogo/io/segio/fai types and values point to this package and note that that package is deprecated since it belongs here more than it does there.

Please take a look.


kortschak:
Note that there is an alternative approach that allows non-file readers to be used and does not have an int-size limitation on 32-bit archs. If the `File` type holds an `io.ReaderAt` instead of opening an `*mmap.ReaderAt`.

kortschak:
@brentp *ping*

kortschak:
i had intended to use nmap, but then decided that `ReaderAt` was more flexible and simpler.

PR #152, opened 2020-10-02
bam: iterate over file when no chunks are given

brentp:
Please take a look.

Previous test was skipped. The `n` incrementing now checks that ieration actually occurs.

### DIFF
--- a/bam/bam_test.go
+++ b/bam/bam_test.go
@@ -422,12 +422,14 @@ func (s *S) TestSpecExamplesIterator(c *check.C) {
 	c.Assert(err, check.Equals, nil)
 	it, err := NewIterator(br, nil)
 	c.Assert(err, check.Equals, nil)
-	for i := 0; it.Next(); i++ {
+	i := 0
+	for ; it.Next(); i++ {
 		expect := specExamples.records[i]
 		r := it.Record()
 		c.Check(r.Name, check.Equals, expect.Name)
 		c.Check(r.Pos, check.Equals, expect.Pos)
 	}
+	c.Assert(i, check.Equals, len(specExamples.records))
 	c.Assert(it.Error(), check.Equals, nil)
 }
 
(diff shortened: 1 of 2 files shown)


bam/reader.go
```diff
@@ -256,7 +256,7 @@ type Iterator struct {
 //
 func NewIterator(r *Reader, chunks []bgzf.Chunk) (*Iterator, error) {
 	if len(chunks) == 0 {
-		return &Iterator{r: r, err: io.EOF}, nil
+		return &Iterator{r: r, err: nil}, nil
```
kortschak, line 259: Just leave the `err` field out and depend on the zero value.

bam/bam_test.go
```diff
@@ -422,12 +422,15 @@ func (s *S) TestSpecExamplesIterator(c *check.C) {
...
 	c.Assert(err, check.Equals, nil)
+	n := 0
 	for i := 0; it.Next(); i++ {
 		expect := specExamples.records[i]
 		r := it.Record()
 		c.Check(r.Name, check.Equals, expect.Name)
 		c.Check(r.Pos, check.Equals, expect.Pos)
+		n += 1
 	}
+	c.Assert(n, check.Equals, 6)
```
kortschak, line 433:
```
	i := 0
	for ; it.Next(); i++ {
		expect := specExamples.records[i]
		r := it.Record()
		c.Check(r.Name, check.Equals, expect.Name)
		c.Check(r.Pos, check.Equals, expect.Pos)
	}
	c.Assert(i, check.Equals, len(specExamples.records))
```

kortschak (changes requested):
Ooops. Thanks for finding this.

bam/bam_test.go
brentp: done

kortschak (approved):
Thanks

PR #155, opened 2021-03-17
bgzf: Added check to avoid nil pointer dereference panic when seeking 

AlanRace:
Occasionally in bgzf/reader.go bg.current is set to nil, which can cause a panic when then attempting to Seek(). This adds a check for nil prior to attempting to use bg.current

### DIFF
--- a/bgzf/reader.go
+++ b/bgzf/reader.go
@@ -449,7 +449,7 @@ func (bg *Reader) Seek(off Offset) error {
 		return ErrNotASeeker
 	}
 
-	if off.File != bg.current.Base() || !bg.current.hasData() {
+	if bg.current == nil || off.File != bg.current.Base() || !bg.current.hasData() {
 		ok := bg.cacheSwap(off.File)
 		if !ok {
 			var dec *decompressor


kortschak:
Can you provide a reproducer please.

AlanRace:
This is something that I have seen, and currently experience. I am using the bgzf code to write a parser for the pairix format (https://github.com/4dn-dcic/pairix) and using the corresponding index file to provide random access to the file. I have been so far unable to write a short test to reproduce the problem, as it seemingly occurs randomly. But once it occurs, subsequent Seeks() also fail.

kortschak:
Thanks. I'd really like to see the code that is causing it.

AlanRace:
If I figure out a series of events that causes it, I'll send over code to reproduce - otherwise I can provide a link to the code once published (or at least once released publicly). In the meantime, this is roughly what the code is doing:

```go
bgzfReader, err := bgzf.NewReader(pairsFile, 0)
// Handle errors etc.

// Generate data query
query := NewQuery("chr1:10000-20000|chr1:10000-20000")

// Search index using query to get the virtual file location (start and end)
start, end := index.Query(query)

// Convert virtual file location to bgzf 
func getBGZFOffset(location uint64) bgzf.Offset {
	blockOffset := uint16(location & 0xFFFF)
	blockAddress := int64(location >> 16)

	return bgzf.Offset{File: int64(blockAddress), Block: uint16(blockOffset)}
}
startOffset := getBGZFOffset(start)
endOffset := getBGZFOffset(end)

// Go to start location
err = file.bgzfReader.Seek(startOffset)
if err != nil {
	// Handle error
}

// Create a buffered reader to read in data line by line
bufReader := bufio.NewReader(file.bgzfReader)

for ; ; {
	lineData, err := bufReader.ReadBytes('\n')
	if err != nil {
		// Handle error
	}

	// Process lineData

	// If we've gone past the end of our query region, then stop
	if bgzfReader.LastChunk().Begin.File > endOffset.File {
		break
	}
}


```

kortschak:
Thanks. If you could email me the necessary files for the reproducer, that would be very helpful.

Note that your use of `bufio.Reader` in conjunction with `r.LastChunk()` there is not safe; it decouples the bgzf reader's position from the user's understanding of the cursor position since the bgzf reader will have read more than has been presented to the user. I doubt this is the cause of the problem, but it is something that you should be aware of. You should use a `bgzf/index.ChunkReader` to get your file chunks.

AlanRace:
Thank you for the advice - I'll look into that. As for the issue, I think it might be related to different go routines trying to Seek() at the same time, which would explain the intermittent appearance of the bug. I had a mutex to prevent that from happening, but found a logic error on my side, and since resolving that I haven't seen the panic again (so far).

From what I can see in the code, the cache is protected by a mutex, but seeking etc. isn't, is that correct?

kortschak:
Yeah, `Seek` is not protected by a mutex. I fact nothing user facing is; in general, the operations provided for users are not meaningful when used in multiple goroutines unless they are grouped and the groups of operations are protected. The same is true for using the index package. Indeed, if you are doing threaded operations on a file you should use more than one Reader, one for each worker. This is due to the use of seek/read rather than pread as the underlying filesystem i/o. The choice of seek/read allows streams and network data to be used at the cost of loss of thread safety. In short, what it sounds like you are doing is probably not a good idea.

If you think it's due to races, run your code with `-race`.

AlanRace:
As I said, I have a mutex to ensure Seek/Read interactions with the file are executed together, there was just a bug in my code meaning that it wasn't locked under certain conditions. I think what I am doing is fine, as I don't want/intend to access the file from multiple go routines at the same time - they can all wait their turn. Multiple readers wouldn't really solve the problem in my opinion either as there is still only a single file being accessed so will still be IO bound.

I am now pretty certain this was related to race conditions, so feel free to close this if you like. Perhaps you could throw an error instead of panicing though? I don't think the inability to seek should crash the program. I'd be happy to write that code if desired.

kortschak:
I didn't see that the seek/read was protected together. But if you're doing that, then that's fine.

If the cause is a race condition, then a panic is the appropriate behaviour.

PR #156, opened 2021-05-22
internal/pool: add a buffer pool and use it

kortschak:
DO NOT MERGE

This exists just to demonstrate that a pool is not effective in case it comes up in the future.

```
name              old time/op    new time/op    delta
Read-8               4.24s ± 3%     4.67s ± 1%  +10.26%  (p=0.000 n=9+10)
ReadCoreAndSeq-8     3.37s ± 6%     3.58s ± 0%   +6.26%  (p=0.000 n=10+9)
ReadCoreOnly-8       3.32s ± 6%     3.52s ± 2%   +6.05%  (p=0.000 n=10+10)
ReadFile-8           4.19s ± 1%     4.66s ± 1%  +11.14%  (p=0.000 n=9+9)

name              old alloc/op   new alloc/op   delta
Read-8              2.35GB ± 0%    2.13GB ± 0%   -9.42%  (p=0.000 n=10+10)
ReadCoreAndSeq-8    1.05GB ± 0%    0.74GB ± 0%  -30.14%  (p=0.000 n=10+10)
ReadCoreOnly-8       991MB ± 0%     673MB ± 0%  -32.09%  (p=0.000 n=9+10)
ReadFile-8          2.35GB ± 0%    2.13GB ± 0%   -9.42%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
Read-8               18.1M ± 0%     24.1M ± 0%  +32.95%  (p=0.000 n=10+10)
ReadCoreAndSeq-8     12.3M ± 0%     16.3M ± 0%  +32.59%  (p=0.000 n=10+10)
ReadCoreOnly-8       10.3M ± 0%     14.3M ± 0%  +38.93%  (p=0.000 n=10+10)
ReadFile-8           18.1M ± 0%     24.1M ± 0%  +32.95%  (p=0.000 n=10+10)
```

PR #157, opened 2021-05-22
bam: use heuristics to reduce reallocation

kortschak:
The only remaining area of optimisation is in decoding headers for the allocation of `text`, but there only pooled buffers are possible. I tried this using the system in #156 and it had no impact either way.

Please take a look.


brentp (approved):
Seems fine to me from what I understand. Does this improve performance measurably?

kortschak:
Yes, it does - benchmark results are in the commit messages.

PR #158, opened 2021-06-01
bgzf: satisfy io.ByteReader

kortschak:
This allows more efficient indexing of tabix formatted or other line-based data.

Please take a look.


kortschak:
Recommend review with whitespace diff turned off to reduce diff noise.

brentp (approved):

Issue #159, opened 2021-06-11
bgzf: Seek() loads wrong block due to readahead caching bug


csw:
### What are you trying to do?

Seek to read records from a tabix-indexed BGZF file with caching and multiple decompressors enabled.

### What did you do?

I've created a [test case](https://github.com/csw/hts/blob/bgzf-seek-cache-fix/bgzf/bgzf_test.go#L1205) that can reproduce this. Note that it's timing-sensitive; without the `time.Sleep()` call, the test will pass on my machine.

### What did you expect to happen?

`Seek()` should position the reader at the specified position in the file, and a subsequent read should return the expected bytes.

### What actually happened?

With a cache, multiple decompressors, and enough wait time for a readahead task to complete before calling `Seek()`, the reader will be positioned at the correct block offset within the wrong block and return the wrong data from a subsequent read. Internally, if the block being seeked to is at the head of the `waiting` queue, it'll be added to the cache [on line 464](https://github.com/biogo/hts/blob/5435c6ae8d4d8bf9b56c3d4a8acc4a8b50a1fe14/bgzf/reader.go#L464), causing `nextBlockAt()` to return the position of the next unread block after the requested block.

I have a fix ready: 34fa03c4522b61a67fb654702f896337376f6d38. I can make a PR for it if it looks good.

### What version of Go and bíogo/hts are you using?

```
go version go1.16.4 linux/amd64
5435c6ae8d4d8bf9b56c3d4a8acc4a8b50a1fe14
github.com/biogo/hts v1.4.0 h1:ArS2PExdQpGnUbLYpfQYXIAZHbMQRQPOlF5TP9xZncE=
github.com/biogo/hts v1.4.0/go.mod h1:3D5xDsTeUqVtI5hFFIAKUL2sUPtXmbRTGyzhfuExWMQ=
```

### Does this issue reproduce with the current master?

Yes.

kortschak:
Thank you for finding this; nice work. Also thank you for working on a fix, it's pretty subtle and I'm glad I didn't have to spend time on it.

Yes, please send a PR. To short circuit comments there, please make the comment in fix, `// This decompressor already had the block we wanted.` and add a comment in the `dec != nil` condition explaining why this work is conditional (the comment above suggests it, but it will be easier in the future if there is an explicit statement).

Can you also send a PR to biogo/biogo "A+C: add Clayton Wheeler" adding yourself to AUTHORS and CONTRIBUTORS?

csw:
[...]

PR #160, opened 2021-06-13
bgzf: fix caching bug in Reader.Seek()

csw:
This fixes biogo/hts#159. When using multiple read decompressors and a cache, and the target block was at the head of the `bg.working` readahead queue, Seek() could return data from the wrong block by populating the cache with the target block and then calling nextBlockAt(), which would return the start of the next unread block rather than the start of the target block. 

I also added a doc comment for `NewReader()` explicitly stating that rd=1 means synchronous reads without readahead, since that wasn't clear to me until I was reading the code.


kortschak (approved):
Thanks

PR #161, opened 2021-07-01
bam: do not share buffer with qual between reads

kortschak:
Please take a look.

Fixes brentp/irelate#5.


brentp (approved):

brentp:
Thank you very much.

kortschak:
No worries, thanks for your help. I was going to suggest you update your go.mod file, but there isn't one, so you don't need to, though probably you should add one.

brentp:
I should. I never kept up with the go modules stuff. Will have to revisit at some point.

PR #165, opened 2023-08-24
Multiple improvements

egonelbre:
While discussing this package I did a quick static analysis check, these are some minor quality improvements:

- bump the language version to Go 1.19, mainly for module handling improvements and some funcs
- fixes to deprecated functions
- use errors.Is for checking for errors instead of `==`
- some automated fixes from golangci and staticcheck
- fixes to error handling
- use t.TempDir instead of os.CreateTemp


bam/add_example_test.go
```diff
@@ -22,7 +23,7 @@ func ExampleIndex_Add() {
 	var bai bam.Index
 	for {
 		r, err := br.Read()
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
kortschak, line 26: `io.EOF` is never wrapped, so I don't think these are necessary.

bam/merger.go
```diff
@@ -126,7 +126,7 @@ func (m *Merger) Read() (rec *sam.Record, err error) {
 func (m *Merger) cat() (rec *sam.Record, err error) {
 	id := m.readers[0].id
 	rec, err = m.readers[0].r.Read()
-	if err == io.EOF && len(m.readers) != 0 {
+	if errors.Is(err, io.EOF) && len(m.readers) != 0 {
```
kortschak, line 129: Same here and throughout. Also for the other errors that are never wrapped.

bam/add_example_test.go
egonelbre: I kind of agree that `io.EOF` should never be wrapped, but I've encountered readers where it happens. Anyways, I can revert this change.

egonelbre: Note, it's usually due to some oversight in a wrapping reader writer that propagates some internal error across some misc. logic.

kortschak (approved):

Issue #167, opened 2024-03-27
fai: Reading a fasta file with blank contig names causes invalid offsets


bloveless:
### What are you trying to do?

When processing a fasta file that has empty contig names I was expecting that the empty names would essentially just be ignored. I noticed that the loop to reset the index record doesn't reset the record [unless the previous contig had a name](https://github.com/biogo/hts/blob/master/fai/fai.go#L63). In my local experiments if I moved the line `rec = Record{}` to be outside of the `if rec.Name != "" {` block then things seemed to work correctly. I'd be happy to submit a PR if this is a desired change.

Alternatively, if this is considered an invalid FASTA file then returning an error upon detecting this would work for me as well.

### What did you do?

In the following example you can see that the first index that is generated has a length in the 200's. The second index all I did was added a name to the first sequence and the length returned to a normal value.

https://go.dev/play/p/f3A7itF0gqb

### What did you expect to happen?

I expected the contig that had no name to be ignored but the offsets and lengths for the next contig to be correct and usable.

### What actually happened?

It seems like the length from the no-name contig is added to the length of the next contig. The start seems to be accurate but the BasesPerLine and BytesPerLine are still held from the no-name contig and the next contig is checked against those values. 

### What version of Go and bíogo/hts are you using?

github.com/biogo/hts v1.4.4 h1:Z+TminqAKRE/t6nyy5PwI/DL90kdew4GpghB+QdjjFk=
github.com/biogo/hts v1.4.4/go.mod h1:AfPn4uJQ2zxi04Q/4vccdmCX16W+IsHXVguPsdh4HE4=

### Does this issue reproduce with the current master?

yes

PR #168, opened 2024-03-27
fai: return error when generating index if an empty or all whitespace contig name is encountered

bloveless:
Resolves: https://github.com/biogo/hts/issues/167

When a sequence name is encountered that is a sole > character or a > with trailing whitespace. Return an error.

Previously this scenario would have continued without error and corrupted the indexes of the next sequence.

As a note: I also encountered an error in the bgzf test when running locally so I fixed that as well. All tests pass when running `go test ./...` locally.

Related authors and contributors PR in accordance with the guidelines. https://github.com/biogo/biogo/pull/90


fai/fai.go
```diff
@@ -57,6 +57,9 @@ func NewIndex(fasta io.Reader) (Index, error) {
 		if len(b) == 0 {
 			continue
 		}
+		if bytes.Equal(bytes.TrimRight(b, " "), []byte{'>'}) {
```
kortschak, line 60: `b` has been space-trimmed, so it should be enough to compare to the ">" literal.

fai/fai_test.go
```diff
@@ -85,6 +85,33 @@ TTCATTAATCATTTGTACTGGATCTGTTCG
...
+GCGAACTGATGGTCAAGCACAGCTAGAACGTCTTCGTTTGAAGCTGGAGACGATTGCCGCGCAGGCAACT
+CGCCGACATCGACGATGTCTTCGTAGTCAT
+>FN654386.3
+CTGAATTGAGCCGCGGGCGATTGATTCGTGTCGGCGCGTCAGGAGGAAGTTCAAGTCGGAATCTCGCGTT
+TTCATTAATCATTTGTACTGGATCTGTTCG
+`,
+			idx: nil,
+			err: errors.New("fai: missing sequence name at 104"),
+		},
+		{
```
kortschak, line 100: Why are there two identical tests? (`idx` is nil in the second implicitly).

bloveless: The second test has only spaces for the name. I'm happy to remove that test though.

bgzf/bgzf_test.go
```diff
@@ -108,12 +108,13 @@ func TestEOF(t *testing.T) {
...
 		// NOTE: on Windows, os.Open is not allowed for directories.
-		f, err = os.Open(t.TempDir())
+		dir := t.TempDir()
+		f, err = os.Open(dir)
 		if err != nil {
 			t.Fatalf("Open temp dir: %v", err)
 		}
 		ok, err = HasEOF(f)
-		if want := "read " + t.TempDir() + ": is a directory"; err.Error() != want {
+		if want := "read " + dir + ": is a directory"; err.Error() != want {
```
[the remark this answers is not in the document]
bloveless: Running the tests locally failed for me because it generated two different tempdirs. None-the-less I will revert.

fai/fai.go
bloveless: Resolved in 25658fd85b0583d425debc45806b89a25d9c61b0

fai/fai_test.go
bloveless: Resolved in 25658fd85b0583d425debc45806b89a25d9c61b0

kortschak: That's pretty brittle and not very clear. If that's the intention, please put back the explicit `idx: nil` so that its not confounding, and include some syntax that shows that there is an explicit set of spaces. Something like
```
>`+"    "+`
```

bloveless: Will do

bloveless: Re-added in 6babe23

kortschak (approved):
Thanks

bloveless:
@kortschak thanks for merging this and for this library in general! Will this change result in a new tag or do you need me to do anything else before that happens?

kortschak:
Push tag v1.4.5.

bloveless:
Awesome! I see the new tag. Thank you.

